PR #8, opened 2017-01-27
Feature/drive to position command

imathno:
Implemented DriveToPositionCommand
Added new method to DriveTestBase
Added test for DriveToPositionCommand


Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -5,6 +5,8 @@
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
+import edu.wpi.first.wpilibj.Talon;
```
WasabiFan: What's this doing? It shouldn't be needed (if it is, something else is wrong).

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -43,6 +49,13 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
         dVelProp = propManager.createPersistentProperty("Drive Vel D", -100);
         fVelProp = propManager.createPersistentProperty("Drive Vel F", 0);
         
+        leftDistanceProp = propManager.createEphemeralProperty("Left Distance", 0);
```
WasabiFan: Capitalize only the first letter of the property string; everything else should be lowercase.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -148,7 +163,6 @@ public void tankDrivePowerMode(double leftPower, double rightPower) {
      * @param motor used to access the motor to set and give data
      */
     private void updateMotorConfig(XCANTalon motor) {
-        motor.configEncoderCodesPerRev((int)encoderCodesProperty.get());
```
WasabiFan, line 151: Why was this removed? Is the idea to do all the conversions ourselves? If so, the unused property should be removed. Also, the value should be explicitly set to 1 so that we aren't using whatever value was last set.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -164,5 +178,20 @@ public void updateMotorTelemetry() {
         leftDriveSlave.updateTelemetryProperties();
         rightDrive.updateTelemetryProperties();
         rightDriveSlave.updateTelemetryProperties();
+        
+        leftDistanceProp.set(leftDrive.getPosition());
+        rightDistanceProp.set(rightDrive.getPosition());
+    }
+    
+    public double getDistance() {
+        return convertToInches(leftDrive.getPosition() - startPosition.get());
```
WasabiFan: It'd feel safer if this used both right and left (i.e. averaged them).

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -164,5 +178,20 @@ public void updateMotorTelemetry() {
...
+    }
+    
+    public void resetDistance() {
+        startPosition.set(convertToInches(leftDrive.getPosition()));
```
WasabiFan: This seems odd; we're using a property to frequently store data for immediate access by other code. I think this value should be stored in a local variable independent of the property.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: `startPosition` is already in inches, while `getPosition` returns ticks... and then it's put through an inch conversion. Something needs to be moved.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -164,5 +178,20 @@ public void updateMotorTelemetry() {
         leftDriveSlave.updateTelemetryProperties();
         rightDrive.updateTelemetryProperties();
         rightDriveSlave.updateTelemetryProperties();
+        
+        leftDistanceProp.set(leftDrive.getPosition());
+        rightDistanceProp.set(rightDrive.getPosition());
+    }
+    
+    public double getDistance() {
```
WasabiFan: This isn't distance, this is displacement. It should be named as such.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    private final PIDManager travelManager;
+    
+    private final DoubleProperty distanceTolerance;
+    private double targetPosition;
+    
+    @Inject
+    public DriveToPositionCommand(DriveSubsystem driveSubsystem, XPropertyManager propManager) {
+        this.driveSubsystem = driveSubsystem;
+        this.requires(driveSubsystem);
+        this.travelManager = new PIDManager("DriveToPosition", propManager, 0.1, 0, 0);
```
WasabiFan: Spaces and lowercase in string so that it's easier to read.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,48 @@
...
+        
+        distanceTolerance = propManager.createPersistentProperty("Distance Tolerance", 0.1);
```
WasabiFan: Wow, that's a pretty tight threshold! This should probably be increased a bit.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+    
+    public void setTargetPosition(double targetPosition) {
+        this.targetPosition = driveSubsystem.convertToInches(targetPosition);
```
WasabiFan: These unit conversions scare me. It doesn't feel natural for the position to be in ticks here (I'd expect everything outside of the drive subsystem to be human units). If we do need to mix units, they should be documented.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public void execute() {
+        double power = travelManager.calculate(targetPosition, driveSubsystem.getDistance());
+        
+        driveSubsystem.tankDrivePowerMode(power, power);
+    }
+    
+    @Override
+    public boolean isFinished() {
+        return travelManager.isOnTarget(distanceTolerance.get());
```
WasabiFan: Again -- are these inches or ticks?

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.google.inject.Inject;
+
+import xbot.common.command.BaseCommand;
+import xbot.common.math.PIDManager;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+import competition.subsystems.drive.DriveSubsystem;
+
+public class DriveToPositionCommand extends BaseCommand{
```
WasabiFan: This feels more like a "drive for distance" command than a "drive to position" command. A _position_ implies to me that it's a point on the field in both dimensions.

Competition2017/tests/competition/subsystems/drive/DriveTestBase.java
```diff
@@ -30,4 +31,14 @@ public void verifyDriveSetpoints(double left, double right) {
         assertEquals(left, ((MockCANTalon)drive.leftDrive).getSetpoint(), 0.001);
         assertEquals(right, ((MockCANTalon)drive.rightDrive).getSetpoint(), 0.001);
     }
+    
+    public void verifyDrivePositive() {
+        assertTrue(((MockCANTalon)drive.leftDrive).getSetpoint() > 0);
```
WasabiFan, line 33: Is there a comparison assertion method? I can't remember if there is with JUnit.

Competition2017/tests/competition/subsystems/drive/commands/DriveToPositionCommandTest.java
```diff
@@ -0,0 +1,47 @@
...
+        verifyDriveNegative();
+    }
+    
+    @Test
+    public void isFinishedTest() {
+        DriveToPositionCommand command = injector.getInstance(DriveToPositionCommand.class);
+       
+        command.setTargetPosition(2);
+        
+        ((MockCANTalon)drive.leftDrive).setPosition(2);
```
WasabiFan: Can we also test something within the tolerance but not exactly at the target?

WasabiFan (dismissed):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: I think @aschokking told me to remove this because they put in another implementation of this somewhere else.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: There is no right talon on the summer bot so I was only using the left.

WasabiFan: I'm not sure what you mean. The summer bot has four drive talons, 2 per side, and one encoder per side.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: Didnt mean to import that.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: Doon

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: Done

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
imathno: I was only getting values from the leftDrive

imathno: startPosition starts at ticks if you go up you'll see there is a doubleProperty

imathno: oops didn't even notice that.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: What is that other implementation? The inch conversion function?

imathno: Idk he just told me to remove this and I think this has nothing to do with the inch conversion function.

WasabiFan: @aschokking can you elaborate?

aschokking: I don't remember anything about this guy, if I had said to remove it that was probably a mistake. 

I agree just leaving it at something like 1.0 would probably be simplest though.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: These were the debug properties that we should probably remove or rename them to be encoder ticks or something

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: Indeed we were only getting values for one of the encoders and it's really tricky to get at the electronics on the summer bot to verify it was really there on the other side and plugged in etc.. As you said we'll want both sides averaged together on the final bot but I imagine we'll be using the summer bot for drive testing at least for the next few meetings. Maybe let's have the code to handle both there but commented out?

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: @WasabiFan are you advocating against having a startPosition property or just to also have a local variable in parallel for reading?

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: Are you saying that there was a technical problem preventing you from getting data? What was the behavior?

You bring up a good point; it would be nice to be able to operate on just a single encoder if something gets disconnected.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
aschokking: Agreed! (should be inches everywhere outside of the drive subsystem)

Competition2017/tests/competition/subsystems/drive/DriveTestBase.java
aschokking: Not that I'm aware of either.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: The latter. I think that the property for telemetry purposes is good; it's just the use of such a property as a local immediate variable that makes me uncomfortable.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: We should check the docs to see if this is a property that is maintained in flash memory. If it is, we need to make sure that we call this everywhere.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
imathno: Done

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
imathno: Updated the value to 1

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
imathno: Everything in the DriveSubsystem is now in ticks

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
imathno: Everything in the DriveToPositionCommand is in inches

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
JohnGilb: Agree. This should be renamed to something like "DriveForDistanceCommand"

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -38,10 +42,17 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
...
-        pVelProp = propManager.createPersistentProperty("Drive Vel P", 2);
-        iVelprop = propManager.createPersistentProperty("Drive Vel I", 0);
-        dVelProp = propManager.createPersistentProperty("Drive Vel D", -100);
-        fVelProp = propManager.createPersistentProperty("Drive Vel F", 0);
+        pVelProp = propManager.createPersistentProperty("Drive vel P", 2);
+        iVelprop = propManager.createPersistentProperty("Drive vel I", 0);
+        dVelProp = propManager.createPersistentProperty("Drive vel D", -100);
+        fVelProp = propManager.createPersistentProperty("Drive vel F", 0);
+        
+        leftDistanceProp = propManager.createEphemeralProperty("Left distance", 0);
```
aschokking: let's rename these guys "Left encoder ticks" instead so we don't get confused about units?

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -54,11 +65,13 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
         configMotorTeam(rightDrive, rightDriveSlave);
         rightDrive.createTelemetryProperties("Right master", propManager);
         rightDriveSlave.createTelemetryProperties("Right slave", propManager);
+
+        resetDistance();
```
aschokking: Let's add a command to reset the distance as well

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -148,7 +161,7 @@ public void tankDrivePowerMode(double leftPower, double rightPower) {
      * @param motor used to access the motor to set and give data
      */
     private void updateMotorConfig(XCANTalon motor) {
-        motor.configEncoderCodesPerRev((int)encoderCodesProperty.get());
+        //motor.configEncoderCodesPerRev((int)encoderCodesProperty.get());
```
aschokking: delete line rather than commenting it out

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
JohnGilb: Given that we need to be deliberate about this when reading through it, we should probably append "Ticks" or "Inches" to any properties or variables that contain distance information.

so, "distanceToleranceInches"

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: @JohnGilb is going to verify in the docs if the value is preserved or not.

JohnGilb (changes requested):
See my comments above.

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
aschokking: It is 'To' an absolute distance though as opposed to a relative one. I can see your point about position vs distance but it this year sans mechanum we'll probably never have anything about x/y driving so probably won't get confused with anything else.

aschokking (approved):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
JohnGilb: My most recent read through of the docs suggests that it is not.

Motion Control Profile (MCP) value are saved locally, and those values are:
P Gain: Kp constant to use when control mode is a closed-loop mode.
I Gain: KI constant to use when control mode is a closed-loop mode.
D Gain: KD constant to use when control mode is a closed-loop mode.
F Gain: KF constant to use when control mode is a closed-loop mode.
I Zone: Integral Zone. When nonzero, Integral Accumulator is automatically cleared
when the absolute value of Closed-Loop Error exceeds it.
(Closed-Loop) Ramp Rate: Ramp rate to apply when control mode is a closed-loop
mode.
Allowable Closed-Loop Error: When Closed-Loop Error’s magnitude is less than this
signal, Integral Accum and motor output are auto-zeroed during closed-loop.
Peak Closed-Loop Output: Caps the maximal or peak motor-output during closed-loop.
Nominal Closed-Loop

this does not appear to include encoder configuration.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -164,5 +177,26 @@ public void updateMotorTelemetry() {
...
+        leftDistanceProp.set(leftDrive.getPosition());
+        rightDistanceProp.set(rightDrive.getPosition());
+    }
+    
+    public double getDisplacement() {
+        /* TODO add more values from other motors for accuracy */
+        return convertTicksToInches(leftDrive.getPosition() - startPositionTicks.get());
+    }
+    
+    public void resetDistance() {
```
aschokking: let's call this resetDisplacement so it's clear they're talking about the same thing if we're going with displacement instead

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -47,10 +47,10 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
         dVelProp = propManager.createPersistentProperty("Drive vel D", -100);
         fVelProp = propManager.createPersistentProperty("Drive vel F", 0);
         
-        leftDistanceProp = propManager.createEphemeralProperty("Left distance", 0);
-        rightDistanceProp = propManager.createEphemeralProperty("Right distance", 0);
+        leftEncoderTicksProp = propManager.createEphemeralProperty("Left encoder ticks", 0);
```
WasabiFan: Given that we'll also have left and right shooters, could we include the word "drive" somewhere in the property name?

imathno: Done

Competition2017/src/competition/subsystems/drive/commands/DriveToPositionCommand.java
```diff
@@ -0,0 +1,52 @@
...
+    private final PIDManager travelManager;
+    
+    private final DoubleProperty distanceToleranceInches;
+    private double targetPosition;
+    
+    @Inject
+    public DriveToPositionCommand(DriveSubsystem driveSubsystem, XPropertyManager propManager) {
+        this.driveSubsystem = driveSubsystem;
+        this.requires(driveSubsystem);
+        this.travelManager = new PIDManager("Drive to position", propManager, 0.1, 0.1, 0.1);
```
JohnGilb: Unless you already know that 0.1 is a good constant for the I term, it's recommended to have it be at 0 (since it can lead to extreme robot behavior).

JohnGilb (approved):

PR #13, opened 2017-01-28
Feature/shooter subsystem

Unagifan:
Shooter System for dual shooters. Created using shooters as independent objects.


Competition2017/src/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -24,6 +30,18 @@ public void setupMyCommands(
     */
     
     @Inject
+    public void setupLauncherCommands(
+            OperatorInterface oi,
+            ShooterSubsystem shooterSubsystem,
+            XPropertyManager propertyManager
+            ) {
+        oi.leftButtons.getifAvailable(6).whenPressed(new IncreaseShooterPowerCommand(shooterSubsystem.getLeftShooter()));
```
WasabiFan: I believe these buttons are the same ones that other PRs are using.

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+    //private final XCANTalon slaveMotor;
+    
+    // output telemetry properties
+    private final DoubleProperty shooterCurrentSpeed;
+    private final DoubleProperty shooterTargetSpeed;
+    private final DoubleProperty shooterOutputPower;
+    private final DoubleProperty shooterTalonError;
+    private final BooleanProperty atSpeedProp;
+    
+    //private final DoubleProperty encoderCodesProperty;
```
WasabiFan: Remove commented-out code (everywhere)

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        //encoderCodesProperty = propManager.createPersistentProperty("Launcher Encoder TPC", 4096);
+
+        p = propManager.createPersistentProperty("FuelLauncher wheel P", 0);
+        i = propManager.createPersistentProperty("FuelLauncher wheel I", 0);
+        d = propManager.createPersistentProperty("FuelLauncher wheel D", 0);
+        f = propManager.createPersistentProperty("FuelLauncher wheel F", 0);
+        
+        // Reading through the manual, it looks like only a few things support human units. It might be easier to just keep everything in native units per 100ms,
+        // which is what the Talon uses for all its calculations.
+        // I call this Ticks per Centisecond, or TPC
```
WasabiFan: Are these centiseconds or deciseconds?

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        shooterSpeedThresh = propManager.createPersistentProperty("FuelLauncher nominal speed thresh (TPC)", 1);
```
WasabiFan: Spaces, first letter caps, others lower.

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        shooterTalonError = propManager.createEphemeralProperty("FuelLauncher Speed error", 0);
+        
+        this.masterMotor = factory.getCANTalonSpeedController(1);
+        //this.slaveMotor = factory.getCANTalonSpeedController(8);
+        initiallyConfigureMotorTeam(masterMotor);
+        masterMotor.createTelemetryProperties("FuelLauncher master", propManager);
+        //slaveMotor.createTelemetryProperties("FuelLauncher slave", propManager);
+    }
+
+    private void initiallyConfigureMotorTeam(XCANTalon master) {
```
WasabiFan: Why "initially"? Would "setup..." work? This seems awkward.

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        /*slave.configNominalOutputVoltage(0,  -0);
+        slave.configPeakOutputVoltage(12, -12);
+        slave.enableLimitSwitches(false, false);
+        
+        slave.setControlMode(TalonControlMode.Follower);
+        slave.set(master.getDeviceID());*/
+    }
+    
+    private void updateMotorConfig(XCANTalon motor) {
+        motor.setP(p.get());
```
WasabiFan: Instead of calling this "configuration", can we call it "control params" or similar to differentiate from one-time set-up?

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        masterMotor.ensureTalonControlMode(TalonControlMode.PercentVbus);
+        masterMotor.set(power);
+    }
+    
+    /**
+     * Returns the last set motor power. Only works if the motor is currently in PercentVbus mode - otherwise, returns 0.
+     * @return
+     */
+    public double getLauncherPower() {
+        if (masterMotor.getControlMode() != TalonControlMode.PercentVbus) {
```
WasabiFan: This default behavior is unexpected. Why not return the power being sent regardless of the mode?

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
```diff
@@ -0,0 +1,185 @@
...
+        // voltage, speed, error. What's the best ordering of these things? 
+        // voltage, error, speed. Ideally they would also be restricted to a constant number of digits
+        // to prevent problems.
+        
+        // In the short term, we can pull log files off the RIO for processing, though it adds in
+        // latency.
+        
+        // Format: time, voltage, error, speed
+        
+        log.info(currentTime + ","
```
WasabiFan: This should be a Boolean property. I'd suggest making it ephemeral so that it is reset after a reboot. We only rarely need this, and this method is currently going to be run constantly because it's part of the default command.

Competition2017/src/competition/subsystems/fuel_launcher/commands/MaintainShooterPowerCommand.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public void initialize() {
+    
+    }
+    
+    @Override
+    public void execute() {
+        sideShooter.updateTelemetry();
```
WasabiFan: All this does is update telemetry? If so, `MaintainShooterPowerCommand` is a misnomer; it's just a no-op WRT shooter power.

Competition2017/src/competition/subsystems/fuel_launcher/commands/SetShooterSpeedCommand.java
```diff
@@ -0,0 +1,31 @@
...
+public class SetShooterSpeedCommand extends BaseCommand {
+    
+    final SideShooterSubsystem sideShooter;
+    final DoubleProperty commandedLauncherSpeed;
+
+    public SetShooterSpeedCommand(SideShooterSubsystem sideShooter, XPropertyManager propertyManager) {
+        this.sideShooter = sideShooter;
+        this.requires(this.sideShooter);
+        
+        commandedLauncherSpeed = propertyManager.createEphemeralProperty("Commanded Launcher Speed", 3100.0);
```
WasabiFan: Caps

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,167 @@
...
+        this.side = side;
+        
+        p = propManager.createPersistentProperty(side + "FuelLauncher wheel P", 0);
+        i = propManager.createPersistentProperty(side + "FuelLauncher wheel I", 0);
+        d = propManager.createPersistentProperty(side + "FuelLauncher wheel D", 0);
+        f = propManager.createPersistentProperty(side + "FuelLauncher wheel F", 0);
+        
+        // Reading through the manual, it looks like only a few things support human units. It might be easier to just keep everything in native units per 100ms,
+        // which is what the Talon uses for all its calculations.
+        // I call this Ticks per Centisecond, or TPC
```
WasabiFan: There's lots of duplicated stuff here (from the fuel launcher) and most refers to the name of the other subsystem. Which is supposed to exist and which should be removed? I'd assume this is the correct one, in which case my comments on the other apply here.

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
JohnGilb: Yes, this is a typo - should be deciseconds (since it's in 100ms increments, which is 1/10th of a second)

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
JohnGilb: setupMotorTeam sounds good.

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
JohnGilb: At the time I didn't know how to get the robot-scaled power when the CANTalon was in Speed mode.

Later I understood that masterMotor.getOutputVoltage() / masterMotor.getBusVoltage() was the better approach, but never came back to this bit of code.

That's the better approach, I think, though we still may need the inversion code.

Competition2017/src/competition/subsystems/fuel_launcher/commands/MaintainShooterPowerCommand.java
JohnGilb: Given how the CANTalons work (no new updates mean just keep doing what you're doing), a no-op is actually a "keep doing what you are doing."  

(As a total strawman, if you saw "DoNothingToShooterCommand" on the SmartDashboard, but saw the shooter spinning rapidly under PID control, you would be rightly confused).

@WasabiFan That said, I'm definitely open to better names for anything - what do you think makes more sense here?

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
WasabiFan: Also, we should normalize to `setUp` instead of `setup`; "set up" is the action while "setup" is a noun.

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/commands/MaintainShooterPowerCommand.java
WasabiFan: The reason that it seems wrong is that it specifically calls out "maintain[ing] power" yet it just allows the motor to keep doing what it's doing, which generally _isn't_ maintaining the power. How about `ContinueShooterActionCommand` or `RunCurrentShooterActionCommand`?

It seems like this functionality might be better-served if we had a command specifically for maintaining speed which explicitly kept the talon performing that action. That feels like a much more natural design to me, as otherwise you're essentially losing direct command of your motors (there isn't any state on our end controlling what it does).

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: Yes I was integrating John's code into a new subsystem which works with the dual shooter design. I have now deleted the FuelLauncher subsystem

Competition2017/src/competition/operator_interface/OperatorCommandMap.java
Unagifan: That doesn't seem like the case currently? There is no other pr's at the moment which conflict with these buttons

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/commands/SetShooterSpeedCommand.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/FuelLauncherSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/fuel_launcher/commands/MaintainShooterPowerCommand.java
Unagifan: Do you have any suggestion regarding the naming @WasabiFan ?

Competition2017/src/competition/operator_interface/OperatorCommandMap.java
WasabiFan: As long as you've merged master recently and the tests don't warn you of double allocation, I'm happy.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+import xbot.common.properties.BooleanProperty;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+public class SideShooterSubsystem extends BaseSubsystem {
+    
+    private static Logger log = Logger.getLogger(ShooterSubsystem.class);
+    
+    private final XCANTalon masterMotor;
+    public final String side;
```
WasabiFan: Is there a design reason that this is a string instead of an `enum`? I'd prefer an `enum` unless there's a particular reason not to.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+    private final DoubleProperty p;
+    private final DoubleProperty i;
+    private final DoubleProperty d;
+    private final DoubleProperty f;
+    
+    private final DoubleProperty shooterSpeedThresh;
+    
+   
+    public SideShooterSubsystem(int motor, String side,WPIFactory factory, XPropertyManager propManager){
+        log.info("Creating" + side + " Shooter");
```
WasabiFan: there should be an additional space here and wherever else you do logging using `side`.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+        this.side = side;
+        
+        p = propManager.createPersistentProperty(side + "Shooter wheel P", 0);
```
WasabiFan: If you use a `PIDManager` you can avoid having to create these properties yourself; it does that for you.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+        i = propManager.createPersistentProperty(side + "Shooter wheel I", 0);
+        d = propManager.createPersistentProperty(side + "Shooter wheel D", 0);
+        f = propManager.createPersistentProperty(side + "Shooter wheel F", 0);
+        
+        // Reading through the manual, it looks like only a few things support human units. It might be easier to just keep everything in native units per 100ms,
+        // which is what the Talon uses for all its calculations.
+        // I call this Ticks per Deciseconds, or TPC
```
WasabiFan: Ticks per Decisecond is TPC?

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+    
+    private void setUpMotorTeam(XCANTalon master) {
+        // TODO: Check faults and voltage/temp/current
+        log.info("Configuring" + side + "motor");
+        // Master config
+        master.setFeedbackDevice(FeedbackDevice.QuadEncoder);
+
+        master.reverseSensor(true);
+        master.setInverted(true);
+        log.info("Configuring voltages");
```
WasabiFan, line 82: We can remove this logging now that the issue that was being tracked down is resolved.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+        master.configNominalOutputVoltage(0,  -0);
+        master.configPeakOutputVoltage(12, -12);
```
WasabiFan: Do we actually want to allow it to run at full speed backward? I'd recommend setting the reverse direction to be lower.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+        motor.setI(i.get());
+        motor.setD(d.get());
+        motor.setF(f.get());
+    }
+    
+    /**
+     * Sets the output power of the shooter directly - no PID of any kind
+     * @param power
+     */
+    public void setLauncherPower(double power) {
```
WasabiFan: Given that this is on a shooter object, I don't think we need to specify `launcher`.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+    public void setLauncherTargetSpeed(double speed) {
+        masterMotor.ensureTalonControlMode(TalonControlMode.Speed);
+        
+        // Update property for dashboard
+        shooterTargetSpeed.set(speed);
+        // Instruct motor about new speed goal
+        masterMotor.set(speed);
+        
+        // If there have been any changes to encoder settings or PID setings, apply them.
+        updateMotorPidValues(masterMotor);
```
WasabiFan: It seems like this should be done _before_ setting the target.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+    
+    public double getSpeed() {
+        return masterMotor.getSpeed();
+    }
+    
+    public boolean isAtSpeed() {
+        return Math.abs(getSpeed() - shooterTargetSpeed.get()) <= shooterSpeedThresh.get();
```
WasabiFan, line 143: This is also something that `PIDManager` does for you.

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+        // voltage, speed, error. What's the best ordering of these things? 
+        // voltage, error, speed. Ideally they would also be restricted to a constant number of digits
+        // to prevent problems.
+        
+        // In the short term, we can pull log files off the RIO for processing, though it adds in
+        // latency.
+        
+        // Format: time, voltage, error, speed
+        if(enableShooterLogging.get()){
+        log.info(currentTime + ","
```
WasabiFan: Indentation

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -0,0 +1,166 @@
...
+     * Updates driver station telemetry about the ShooterSubsystem. Try to call this once per scheduler pass-through.
+     */
+    public void updateTelemetry() {
+        masterMotor.updateTelemetryProperties();        
+        atSpeedProp.set(isAtSpeed());
+        shooterCurrentSpeed.set(getSpeed());
+        shooterOutputPower.set(masterMotor.getOutputVoltage() / masterMotor.getBusVoltage());
+        shooterTalonError.set(masterMotor.getClosedLoopError());
+        
+        double currentTime = Timer.getFPGATimestamp();
```
WasabiFan: This var and comment block should go in the `if`.

Competition2017/src/competition/subsystems/shooter/commands/DecreaseShooterPowerCommand.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public DecreaseShooterPowerCommand(SideShooterSubsystem sideShooter) {
+        this.sideShooter = sideShooter;
+        this.requires(this.sideShooter);
+    }
+
+    @Override
+    public void initialize() {
+        log.info("Decreasing" + sideShooter.side + "Shooter's Power");
+        sideShooter.setLauncherPower(sideShooter.getLauncherPower()-0.025);
```
WasabiFan: Magic numbers! I get the feeling that this should be a property.

Competition2017/src/competition/subsystems/shooter/commands/DecreaseShooterPowerCommand.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+    public DecreaseShooterPowerCommand(SideShooterSubsystem sideShooter) {
+        this.sideShooter = sideShooter;
+        this.requires(this.sideShooter);
+    }
+
+    @Override
+    public void initialize() {
+        log.info("Decreasing" + sideShooter.side + "Shooter's Power");
```
WasabiFan: You're going to need to add spaces. Also, caps.

Competition2017/src/competition/subsystems/shooter/commands/IncreaseShooterPowerCommand.java
```diff
@@ -0,0 +1,32 @@
+package competition.subsystems.shooter.commands;
+
+import org.apache.log4j.Logger;
+
+import competition.subsystems.shooter.SideShooterSubsystem;
+import xbot.common.command.BaseCommand;
+
+public class IncreaseShooterPowerCommand extends BaseCommand {
```
WasabiFan: What if these two commands were combined into a "IncrementShooterPowerCommand" or similar that took a parameter for the amount?

Competition2017/src/competition/subsystems/shooter/commands/MaintainShooterPowerCommand.java
```diff
@@ -0,0 +1,25 @@
+package competition.subsystems.shooter.commands;
+
+import competition.subsystems.shooter.SideShooterSubsystem;
+import xbot.common.command.BaseCommand;
+
+public class MaintainShooterPowerCommand extends BaseCommand {
```
WasabiFan: Mentioned previously; rename. Preferably something like `ContinueShooterActionCommand`.

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: Uh I can convert it to a enum type. I didn't do it previously bc I didn't know what a enum was

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/commands/DecreaseShooterPowerCommand.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/commands/MaintainShooterPowerCommand.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
JohnGilb: Sounds like a good protection! If we do that, let's surround that code with a big comment in case we start wondering "why the heck is the power so low?"

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
JohnGilb: Since these values are being set into the CANTalon, it doesn't seem appropriate to use the PIDManager - especially since there's no way today to p/i/d values out of it.

This does suggest that we should build a PIDManagerLite that is just a useful container though!

Competition2017/src/competition/subsystems/shooter/commands/DecreaseShooterPowerCommand.java
Unagifan: !

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
WasabiFan: My thinking was that we could separate the `PIDManager` into a more generic implementation and then use it here.

Competition2017/src/competition/subsystems/shooter/commands/IncreaseShooterPowerCommand.java
Unagifan: !

Unagifan:
@WasabiFan

Unagifan:
@JohnGilb

Competition2017/src/competition/subsystems/shooter/SideShooterSubsystem.java
```diff
@@ -48,14 +53,14 @@ public SideShooterSubsystem(int motor, String side,WPIFactory factory, XProperty
         // Reading through the manual, it looks like only a few things 
         // support human units. It might be easier to just keep everything in native units per 100ms,
         // which is what the Talon uses for all its calculations.
-        // I call this Ticks per Deciseconds, or TPC
+        // I call this Ticks per Deciseconds, or TPD
         shooterSpeedThresh = propManager.createPersistentProperty(side + "Shooter nominal speed thresh (TPC)", 1);
         shooterCurrentSpeed = propManager.createEphemeralProperty(side + "Shooter current speed (TPC)", 0);
```
WasabiFan: These acronyms need to be updated as well

Competition2017/.classpath
```diff
@@ -12,5 +12,9 @@
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib.jar" sourcepath="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="var" path="wpilib" sourcepath="wpilib.sources"/>
```
WasabiFan: Are we _sure_ this is what we want?

WasabiFan (approved):
Small comments; other than that, this looks good!

Competition2017/.classpath
aschokking: eclipse auto generates these, we should avoid checking them in, maybe adding a checkstyle rule would help guard things

WasabiFan: Are you sure? I've been working with this project and AFAIK it hasn't don't this... same with some other Eclipse projects that have a similar JAR configuration. Is there a specific action that causes it?

PR #14, opened 2017-02-01
Feature/driving to orientation command

SethNguyenXBOT:


Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
```diff
@@ -0,0 +1,60 @@
+package competition.subsystems.drive.commands;
+
+import com.google.inject.Inject;
+import xbot.common.command.BaseCommand;
+import xbot.common.math.ContiguousHeading;
+import xbot.common.math.PIDManager;
+import xbot.common.properties.XPropertyManager;
+import competition.subsystems.drive.DriveSubsystem;
+
+public class DriveToOrientationCommand extends BaseCommand{
```
WasabiFan: I think this would be clearer if it were called `RotateToOrientationCommand` or `RotateToHeadingCommand`.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
```diff
@@ -0,0 +1,60 @@
...
+    private XPropertyManager propMan;
+    private PIDManager headingDrivePid;
+    
+    @Inject
+    public DriveToOrientationCommand(DriveSubsystem driveSubsystem, XPropertyManager propMan) {
+
+        this.drive = driveSubsystem;
+        this.propMan = propMan;
+        
+        headingDrivePid = new PIDManager("HeadingModule", propMan, defaultPValue, 0, 0);
```
WasabiFan: That string name isn't very clear. Can we choose a more descriptive name (with spaces) for this?

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public void setHeadings(double heading) {
```
WasabiFan: This should be singular: `setHeading` or `setTargetHeading`.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
```diff
@@ -0,0 +1,60 @@
...
+        
+        targetHeading = new ContiguousHeading(heading);
+    }
+
+    public void setPIDPropertyName(String propertyName){
+
+        headingDrivePid = new PIDManager(propertyName, propMan, defaultPValue, 0, 0);
```
WasabiFan: Why do we need this method? There are some issues that spring out of this but I don't think it is even necessary.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void initialize() {
+        
+    }
+
+    @Override
+    public void execute() {
+        
+        double errorInDegrees = drive.imu.getYaw().difference(targetHeading);
+        double normalizedError = errorInDegrees/180;
```
WasabiFan: What's the goal with normalizing the error?

Competition2017/tests/competition/subsystems/drive/commands/DriveToOrientationCommandTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    public void testTurnLeft90() {
+ 
+        setMockGyro(0);
+        command.setHeadings(90);
+ 
+        command.initialize();
+        command.execute();
+        
+        System.out.println("Left Power: " + ((MockCANTalon)drive.leftDrive).getSetpoint());
```
WasabiFan: It'd be preferable to remove these print calls so that we aren't putting more stuff into the logs than necessary.

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
JohnGilb: Agreed - RotateToHeadingCommand sounds really good.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
JohnGilb: And based on the command rename, it could be something like "Rotate to heading", which would generate "Rotate to heading P", "Rotate to heading I", etc...

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
JohnGilb: agreed, and setTargetHeading sounds very descriptive.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
JohnGilb: Yes, this whole method should be removed - accidentally calling it at the wrong time could be troublesome.

Competition2017/src/competition/subsystems/drive/commands/DriveToOrientationCommand.java
JohnGilb: This was copied from my old code - and I can't remember why I did this. It's probably actually easier to guess at initial PID values without the normalization, since (thanks to geometry classes) we have a better grasp of what 30 degrees is rather than "fractions of a half-circle" which this normalization does.

Long story short - we should remove the normalization.

JohnGilb (changes requested):
see comments

Competition2017/src/competition/subsystems/drive/commands/RotateToHeading.java
```diff
@@ -7,7 +7,7 @@
 import xbot.common.properties.XPropertyManager;
 import competition.subsystems.drive.DriveSubsystem;
 
-public class DriveToOrientationCommand extends BaseCommand{
+public class RotateToHeading extends BaseCommand{
```
JohnGilb: This needs to be RotateToHeadingCommand - all our commands end with the word "Command"

Competition2017/tests/competition/subsystems/drive/commands/RotateToHeadingTest.java
```diff
@@ -7,30 +7,27 @@
 import competition.subsystems.drive.DriveTestBase;
 import xbot.common.controls.actuators.MockCANTalon;
 
-public class DriveToOrientationCommandTest extends DriveTestBase {
+public class RotateToHeadingTest extends DriveTestBase {
```
JohnGilb: Same here - "RotateToHeadingCommandTest"

JohnGilb (changes requested):

Competition2017/src/competition/subsystems/drive/commands/RotateToHeading.java
```diff
@@ -0,0 +1,51 @@
...
+
+    DriveSubsystem drive;
+    ContiguousHeading targetHeading;
+    ContiguousHeading currentHeading;
+    
+    public final double defaultPValue = 1/80d;
+    double headingError;
+    
+    private PIDManager headingDrivePid;
```
aschokking: might as well make this final

Competition2017/src/competition/subsystems/drive/commands/RotateToHeading.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.inject.Inject;
+import xbot.common.command.BaseCommand;
+import xbot.common.math.ContiguousHeading;
+import xbot.common.math.PIDManager;
+import xbot.common.properties.XPropertyManager;
+import competition.subsystems.drive.DriveSubsystem;
+
+public class RotateToHeading extends BaseCommand{
+
+    DriveSubsystem drive;
```
aschokking: final

Competition2017/src/competition/subsystems/drive/commands/RotateToHeading.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void execute() {
+        
+        double errorInDegrees = drive.imu.getYaw().difference(targetHeading);
+        double rotationalPower = headingDrivePid.calculate(0, errorInDegrees);
+
+        drive.tankDrivePowerMode(rotationalPower, -rotationalPower);
+    }
```
aschokking: this command should isFinished when it's stable at it's target heading, otherwise we can't use it in auto to chain things.

Competition2017/tests/competition/subsystems/drive/commands/DriveToOrientationCommandTest.java
aschokking: and better yet should be using the logger instead of print statements so we could control what gets shown/hidden with filter configs.

Competition2017/src/competition/subsystems/drive/commands/RotateToHeading.java
```diff
@@ -0,0 +1,51 @@
...
+import competition.subsystems.drive.DriveSubsystem;
+
+public class RotateToHeading extends BaseCommand{
+
+    DriveSubsystem drive;
+    ContiguousHeading targetHeading;
+    ContiguousHeading currentHeading;
+    
+    public final double defaultPValue = 1/80d;
+    double headingError;
```
WasabiFan: It doesn't look like this is used.

Competition2017/tests/competition/subsystems/drive/DriveTestBase.java
```diff
@@ -27,4 +28,8 @@ public void verifyDriveSetpoints(double left, double right) {
         assertEquals(left, ((MockCANTalon)drive.leftDrive).getSetpoint(), 0.001);
         assertEquals(right, ((MockCANTalon)drive.rightDrive).getSetpoint(), 0.001);
     }
+    
+    public void setMockGyro(double heading) {
```
WasabiFan: How about `setRobotHeading`? The fact that it's a mock gyro isn't important to the caller, given that we only use gyros.

Competition2017/tests/competition/subsystems/drive/commands/RotateToHeadingTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Before
+    public void setup() {
+        
+        super.setUp();
+        
+        command = injector.getInstance(RotateToHeading.class);
+    }
+    
+    @Test
+    public void testTurnLeft90() {
```
WasabiFan: How about something testing that it finishes?

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/drive/commands/RotateToHeadingCommand.java
```diff
@@ -0,0 +1,56 @@
...
+        double errorInDegrees = drive.imu.getYaw().difference(targetHeading);
+        double rotationalPower = headingDrivePid.calculate(0, errorInDegrees);
+
+        drive.tankDrivePowerMode(rotationalPower, -rotationalPower);
+    }
+    
+    @Override
+    public boolean isFinished() {
+        
+        return drive.imu.getYaw() == targetHeading;
```
aschokking: Still need to update this, use `headingDrivePid.isOnTarget()`

aschokking (changes requested):

Competition2017/src/competition/subsystems/drive/commands/RotateToHeadingCommand.java
```diff
@@ -0,0 +1,56 @@
...
+        double errorInDegrees = drive.imu.getYaw().difference(targetHeading);
+        double rotationalPower = headingDrivePid.calculate(0, errorInDegrees);
+
+        drive.tankDrivePowerMode(rotationalPower, -rotationalPower);
+    }
+    
+    @Override
+    public boolean isFinished() {
+        
+        return headingDrivePid.isOnTarget(1);
```
aschokking: make this tolerance a `DoubleProperty` please so we can tune it more easily.

Competition2017/src/competition/subsystems/drive/commands/RotateToHeadingCommand.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public void setTargetHeading(double heading) {
+        
+        targetHeading = new ContiguousHeading(heading);
+    }
+    
+    public void reset(){
+        
+        headingDrivePid.reset();
```
aschokking: who calls reset? Should this happen in initialize probably?

aschokking (changes requested):

Competition2017/src/competition/subsystems/drive/commands/RotateToHeadingCommand.java
JohnGilb: Done

Competition2017/src/competition/subsystems/drive/commands/RotateToHeadingCommand.java
JohnGilb: Using new isOnTarget now.

JohnGilb (approved):

Competition2017/tests/competition/subsystems/drive/commands/RotateToHeadingCommandTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    public void testTurnRight90From150() {
+ 
+        setRobotHeading(150);
+        command.setTargetHeading(90);
+ 
+        command.initialize();
+        command.execute();
+        
+        System.out.println("Left Power: " + ((MockCANTalon)drive.leftDrive).getSetpoint());
```
WasabiFan: This should be removed, as the tests already assert the proper behavior.

JohnGilb: cleaning up

aschokking (approved):

WasabiFan:
Uh... this was merged without it being updated to use the pose system. This code needs to be changed to use it in a separate PR.

aschokking:
Agreed, separate task for the posesubsystem integration: https://app.asana.com/0/38541457243752/262705727435802

PR #16, opened 2017-02-03
Update drive subsystem to use new telemetry source

aschokking:
Wrapped the existing `updateMotorTelemetry()` since there are some other outstanding changes to it in PR and wanted to minimize merge conflicts.


Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -201,4 +202,9 @@ public double convertTicksToInches(double ticks) {
     public double convertInchesToTicks(double inches) {
         return inches * ticksPerInch.get();
     }
+
+    @Override
+    public void updateTelemetry() {
+        this.updateMotorTelemetry();
```
WasabiFan: Why don't we move the contents of that method here?

JohnGilb: Yeah, too many redirects =]

JohnGilb (approved):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: I'd made a comment about this at the top of the PR:
Wrapped the existing updateMotorTelemetry() since there are some other outstanding changes to it in PR and wanted to minimize merge conflicts.

I think the dependent changes have gone in though since I created this PR so can do the refactoring now.

PR #21, opened 2017-02-04
Feature/drive straight

SethNguyenXBOT:


Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -39,6 +42,7 @@
     @Inject
     public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
         log.info("Creating DriveSubsystem");
+        imu = factory.getGyro(ImuType.navX);
```
aschokking: we'll soon replace this with the PoseSubsystem so wouldn't sweat any tweaks to the access of the gyro here.

aschokking (approved):

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
```diff
@@ -26,14 +27,20 @@
...
     private int onTargetCount = 0;
     
     @Inject
     public DriveForDistanceCommand(DriveSubsystem driveSubsystem, XPropertyManager propManager) {
         this.driveSubsystem = driveSubsystem;
         this.requires(driveSubsystem);
         this.travelManager = new PIDManager("Drive to position", propManager, 0.1, 0, 0, 0.5, -0.5);
-        
+
+        headingDrivePid = new PIDManager("HeadingModule", propManager, defaultPValue, 0, 0);
```
WasabiFan: Spaces, caps

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
```diff
@@ -26,14 +27,20 @@
...
+        targetHeading = new ContiguousHeading();
         onTargetCountThresholdProp = propManager.createPersistentProperty("DrvToPos min stabilization time", 3);
```
WasabiFan: This property name should be clarified.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
```diff
@@ -71,7 +80,21 @@ public void initialize() {
...
+        driveSubsystem.tankDrivePowerMode(leftPower, rightPower);
+    }
+    
+    public double calculateHeadingPower() {
+
+        double errorInDegrees = driveSubsystem.imu.getYaw().difference(targetHeading);
+        double normalizedError = errorInDegrees / 180;
+        double rotationalPower = headingDrivePid.calculate(0, normalizedError);
+
+        rotationalPower *= -1;
```
WasabiFan: Instead of negating it here, why not swap the operands in the `difference` call?

WasabiFan (changes requested):
We need to make sure that this isn't merged before it is updated to use the `PoseSubsystem`.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
aschokking: There was definitely a lot of back and forth on something descriptive but not 10 words long here, open to suggestions.

WasabiFan: Most importantly, this should legitimately be time, instead of number of calls to `isFinished`.

Naming-wise, `Stopped duration for completion` would be a reasonable phrase.

aschokking: I'm not sure it's the **most** important thing to make it time, though I agree it would be more robust in theory. The loop counts in practice have worked very well and is a lot simpler to initially write and validate the logic thereof (came about while testing the command and having very poor accuracy when the command completed). Case in point, our current PID logic has no concept of time even though in theory it should to accurately compute the derivative and accumulated errors for instance.

There's a task to fold the concept of 'on target for time' into the PIDManager and I agree it probably makes sense to have it to be `time` based there.

WasabiFan: In general, the contract when defining an `execute` method is that it will be called consistently, once per loop; that makes the current PID implementation work, although it would preferably be improved. On the other hand, there's no such expectation with `isFinished`: it's a getter, and getters should never have side-effects. If it does, it's no longer a getter. This is most impactful in tests, where we are manually calling these methods. In a test case, it makes a lot more sense to a reader to call it once and assert, change the elapsed time, and assert again than to repeatedly call the `isFinished` method until it changes its mind.

If this were inside the `PIDManager` or the `PID` class, I'd stick with number of calls to `execute` as that is what the rest of the PID logic uses. However, here, this has no attachment to the mechanisms of the PID logic: it should use time if that's what makes the most sense in the context of its task.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
JohnGilb: done

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
JohnGilb: Sure

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceCommand.java
JohnGilb: This seems like a reasonable solution for a student, and should serve us well enough until we have a more robust time-stabilization method in place at the PIDManager level.

PR #20, opened 2017-02-04
Created a ShooterFeeder subsystem, commands, and tests.

SethNguyenXBOT:

PR #22, opened 2017-02-05
Added a new subsystem for the rope aligner

imathno:


Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,35 @@
...
+
+import xbot.common.command.BaseSubsystem;
+import xbot.common.controls.actuators.XCANTalon;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class RopeAlignerSubsystem extends BaseSubsystem {
+    private static Logger log = Logger.getLogger(ClimbingSubsystem.class);
```
JohnGilb: `getLogger(RopeAlignerSubsystem.class)`

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,35 @@
...
+    
+    public final XCANTalon intakeMotor;
+    private final DoubleProperty powerToMotor;
+    
+    @Inject
+    public RopeAlignerSubsystem(WPIFactory factory, XPropertyManager propManager) {
+        log.info("Creating Rope Aligner Subsystem");
+        this.intakeMotor = factory.getCANTalonSpeedController(63);
+        this.powerToMotor = propManager.createPersistentProperty("Align motor power", 0.1);
```
JohnGilb: `Rope aligner motor power`

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    
+    public void intake() {
+        intakeMotor.set(powerToMotor.get());
+    }
+    
+    public void stopIntake() {
+        intakeMotor.set(0);
+    }
```
JohnGilb, line 38: you should also have an "eject".

Competition2017/src/competition/subsystems/climbing/commands/RopeAlignerCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    
+    @Inject
+    public RopeAlignerCommand(RopeAlignerSubsystem ropeAlignerSubsystem) {
+        this.ropeAlignerSubsystem = ropeAlignerSubsystem;
+        this.requires(ropeAlignerSubsystem);
+    }
+    
+    @Override
+    public void initialize() {
+
```
JohnGilb: Commands should log that they are being initialized. See other commands for examples

JohnGilb (changes requested):
You also need to add these commands to the command map, and add a default command.

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
imathno: OOPS

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
imathno: done

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
imathno: done

Competition2017/src/competition/subsystems/climbing/commands/RopeAlignerCommand.java
imathno: doon

Competition2017/src/competition/subsystems/climbing/commands/RopeAlignerCommand.java
```diff
@@ -17,7 +20,7 @@ public RopeAlignerCommand(RopeAlignerSubsystem ropeAlignerSubsystem) {
     
     @Override
     public void initialize() {
-        
+        log.info("Initializing RopeAlignerCommandS");
```
JohnGilb: extra "S"

JohnGilb (changes requested):
one comment

Competition2017/src/competition/subsystems/climbing/commands/RopeAlignerCommand.java
imathno: When you're trying to save all

JohnGilb (approved):

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+import xbot.common.controls.actuators.XCANTalon;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class RopeAlignerSubsystem extends BaseSubsystem {
+    private static Logger log = Logger.getLogger(RopeAlignerSubsystem.class);
+    
+    public final XCANTalon intakeMotor;
```
WasabiFan, line 18: This should be private.

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    private final DoubleProperty powerToMotor;
+    
+    @Inject
+    public RopeAlignerSubsystem(WPIFactory factory, XPropertyManager propManager) {
+        log.info("Creating Rope Aligner Subsystem");
```
WasabiFan, line 23: Caps

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
```diff
@@ -0,0 +1,39 @@
...
+    
+    public void intake() {
+        intakeMotor.set(powerToMotor.get());
+    }
+    
+    public void eject() {
+        intakeMotor.set(-powerToMotor.get());
+    }
+    
+    public void stopIntake() {
```
WasabiFan, line 36: It would be clearer if this were called just "stop": otherwise it makes it seem like it's only for `intake()`.

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
aschokking: We're supporting this model to make writing tests lighter weight initially. We'll police bad behavior (anyone else referencing the property than the unit tests) at the PR level and if there's extra time people can definitely go in and refactor things with the extra classes if they want to after the fact.

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
aschokking: my preference would be for these kinds of log messages to reference the class name verbatim `RopeAlignerSubsystem`

Competition2017/src/competition/subsystems/climbing/RopeAlignerSubsystem.java
WasabiFan: I don't see your point. We can either _hope_ it's safe or add the extra class to _make_ it safe; the latter has been done in other PRs and is working well with very little effort, so it doesn't make much sense to abandon that now.

aschokking: I agree the extra class _makes_ it safe, but I disagree that its "very little effort".  I personally find the extra classes cumbersome and too enterprise-fizz-buzz-ey for the complexity of code we're writing. That said, if people want to go that route I won't stand in anyone's way but I support the option not to have to write that extra boiler plate code for simple systems.

WasabiFan: It seems the opposite to me. We need to protect ourselves from this because we have a bunch of new programmers who could make the mistake of accessing this field, and have demonstrated that we don't catch all of these issues in review.

Importantly, the work to add these classes is minimal: it's so easy that writing it yourself is less work than copy-pasting. The most recent implementation of this sort is [here](https://github.com/Team488/TeamXbot2017/pull/20/files#diff-58e80124d733054e49f238191cb87871) -- one class with the default constructor (Eclipse does that for you) and a method per property.

JohnGilb: I concur with Alex's general point. Recent bots used public actuators/sensors, Earlier robots used a global map of all actuators/sensors. In all cases, we didn't get into too much trouble that I can remember.

Since your initial advocacy of this approach, when students ask how to start writing tests, I present them with two options - the public variable or the wrapper class, and talk a bit about the merits/disadvantages of both. Students naturally gravitate to the approach that makes more sense to them at the time - and thus, the one they're going to write/debug with more proficiency.

WasabiFan: Last year, I remember two cases where public members were used instead of the proper access methods (I believe one was with the arm motors and the other was some variety of limit switch); the year before, there was one case which spread throughout multiple subsystems. In those three cases, I was the one who fixed the logic to use the correct accessors. As I mentioned above, we haven't previously been catching these issues and this season there have been multiple problems that slipped past review as well. Will this particular instance be the one that is mis-used? Probably not. But there's a worryingly good chance that one of these cases where we're leaving things exposed out of simplicity will come back to bite us.

> Since your initial advocacy of this approach, when students ask how to start writing tests, I present them with two options - the public variable or the wrapper class, and talk a bit about the merits/disadvantages of both. Students naturally gravitate to the approach that makes more sense to them at the time - and thus, the one they're going to write/debug with more proficiency.

That's an error of method, not commentary on the idea. By the logic presented here, we should use an iterative robot next year and program it in Scratch.

PR #25, opened 2017-02-08
Renaming of Agitator to Eggbeater subsystem/commands

aschokking:


WasabiFan:
Why was this name chosen? To me, this makes it _less clear_: anyone vaguely familiar with robotics (and many who aren't) would understand the mechanism "agitator", while it's unlikely that someone would understand the highly Americanized "eggbeater".

aschokking:
Good question, we had a meeting with the other teams with the goal of coming up with a set of mutually agreed upon names for each of the primary mechanisms on the robot (since everyone had been referring to them as different things). Personally I agree that _agitator_ is better but the consensus view there was that _eggbeater_ was more explicit and at this point it's more important that everyone be using the same terms than we rogue off with what we think is best :)

JohnGilb (approved):
Looks like a fairly simple rename.

WasabiFan:
Uniformity is good, but this is pretty egregious; at this point we should be prioritizing keeping our code readable over using the verbal name.

aschokking:
I appreciate the feedback. If you feel very passionate about it I'd encourage you to work with the other teams to agree upon what you would consider a better name and I will be happy to do another refactor to update the code at that time.

WasabiFan:
Great! I've talked it over with members of other teams and concluded that "agitator" should be the term that we use. While this particular design of the agitator arms is called an 'eggbeater", the subsystem for general use is the agitator.

PR #26, opened 2017-02-08
Rename Feeder to ShooterBelt

aschokking:
Continuing saga in unifying the code with the agreed upon mechanism terms with the rest of the team. Feeder is ambiguous in a lot of ways because the eggbeaters feed balls to the belts which feed balls to the shooter wheels. 

Still need to come up with a better name for the parent class that manages the left/right versions of things but went John's initial suggestion of ShooterBeltMasterSubsystem in the short-run (what's there now is worse so this is still a net improvement in clarity though I think there's room to do better).


JohnGilb (approved):
`ShooterBeltFactory` ? =]

PR #39, opened 2017-02-11
Runs belt if shooter wheel is at target speed.

SethNguyenXBOT:


JohnGilb:
This looks like the wrong branch was pushed or something.

PR #44, opened 2017-02-11
Added so that the drive code uses the second encoder

imathno:


aschokking (approved):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -189,6 +184,10 @@ public double convertTicksToInches(double ticks) {
     public double convertInchesToTicks(double inches) {
         return inches * ticksPerInch.get();
     }
+    
+    public double getPositionAverageTicks() {
```
WasabiFan, line 188: This should be "distance" instead of "position"

PR #47, opened 2017-02-12
Feature/tank drive with gamepad

Unagifan:
Completed implementation for TankDrive with Gamepad. Successfully Tested on 2017 Practice Chassis. Created task on Asana with finding a way to invert on the Gamepad buttons. Inverted the leftDrive of Robot as it was previously going in the wrong direction.


aschokking (approved):
Bonus points for a simple test case

PR #51, opened 2017-02-12
Refactor 'sided' command creation

aschokking:
* Created a 'Base' command for each subsystem to capture common code like having the subsystem as a class variable and calling requires
* Updated ShooterBelt and ShooterWheel commands to be consistent with Agitator where the command will be passed in it's subsystem in the constructor instead of having the setter method. The reason for this change was many times we had forgotten when setting up commands to call the setter and this way it will be enforced at compile time. 
* If the commands end up needing more and more dependencies to the point where this approach becomes a problem we can update things to use a factory to create commands to insulate callers from any injectable dependencies.


Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltIfWheelAtSpeedCommand.java
```diff
@@ -4,20 +4,18 @@
...
-        this.beltSubsystem = beltSubsystem;
-        this.requires(beltSubsystem);
-        targetBeltSpeed = beltSubsystem.getTargetSpeed();
+    protected final ShooterWheelSubsystem shooterWheelSubsystem;
+    private double targetBeltSpeed;
+
+    public RunBeltIfWheelAtSpeedCommand(ShooterBeltSubsystem shooterBeltSubsystem, ShooterWheelSubsystem shooterWheelSubsystem) {
+        super(shooterBeltSubsystem);
+        this.shooterWheelSubsystem = shooterWheelSubsystem;
+        targetBeltSpeed = shooterBeltSubsystem.getTargetSpeed();
```
JohnGilb, line 18: This just gets the most recent set target speed, which could be anything (e.g. 0). This needs to read directly from some new DoubleProperty on the BeltSubsystem like `BeltIntakeSpeed`

JohnGilb: And you don't need it here, you could just read that in the `execute()`.

Competition2017/src/competition/subsystems/shooter_belt/commands/RunShooterBeltCommand.java
```diff
@@ -0,0 +1,26 @@
...
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing " +  this.getName());
+    }
+
+    @Override
+    public void execute() {
+        shooterBeltSubsystem.setTargetSpeed(targetBeltSpeed);
+        shooterBeltSubsystem.updatePeriodicData();
```
JohnGilb: We can delete this now that the shooterBelt is a PeriodicDataSource

Competition2017/src/competition/subsystems/shooter_wheel/commands/RunShooterCommand.java
```diff
@@ -33,8 +23,8 @@ public void initialize(){
     
     @Override 
     public void execute(){
-        shooterWheel.setTargetSpeed(commandedLauncherSpeed.get());
-        shooterWheel.updatePeriodicData();
+        shooterWheelSubsystem.setTargetSpeed(commandedLauncherSpeed.get());
+        shooterWheelSubsystem.updatePeriodicData();
```
JohnGilb: as above, remove this

JohnGilb (approved):

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltIfWheelAtSpeedCommand.java
aschokking: https://app.asana.com/0/38541457243752/270337401932322 should fix this, agreed it's bad

PR #52, opened 2017-02-12
Feature/cleanup base xcan talon speed controlled subsystem

JohnGilb:


Competition2017/src/competition/subsystems/BaseXCANTalonPairSpeedControlledSubsystem.java
```diff
@@ -0,0 +1,44 @@
...
+            WPIFactory factory,
+            PIDPropertyManager pidPropertyManager,
+            XPropertyManager propManager) {
+        super(name, masterChannel, invertMaster, invertMasterSensor, factory, pidPropertyManager, propManager);
+        
+        followerMotor = factory.getCANTalonSpeedController(followChannel);
+        initializeFollowerMotorConfiguration(invertFollower);
+    }
+    
+    protected void initializeFollowerMotorConfiguration(boolean motorInverted) {
```
aschokking, line 38: is there a reason not to just do this in the constructor?

aschokking (approved):

Competition2017/src/competition/subsystems/BaseXCANTalonPairSpeedControlledSubsystem.java
WasabiFan: I'd prefer to not do any external communication in the ctor. This way, you can quickly create one of these objects and then actually do the synchronous initialization in a proper initialization context.

WasabiFan (approved):

Competition2017/src/competition/subsystems/BaseXCANTalonPairSpeedControlledSubsystem.java
JohnGilb: My initial thought was that somebody might want to override the config and modify/add their own extra config, so keeping it separate makes that easier to scope your override.

PR #53, opened 2017-02-15
Tidy up some code

imathno:
Added logger to more classes
Removed unecessary imports


Competition2017/src/competition/subsystems/agitator/AgitatorSubsystem.java
```diff
@@ -30,7 +32,7 @@
     private final PIDPropertyManager pidProperties;
 
     public AgitatorSubsystem(int motor, RobotSide side, WPIFactory factory, XPropertyManager propManager, RobotAssertionManager assertionManager){
-        log.info("Creating " + side + " Agitator");
```
aschokking: In order for it to work removing the 'side' here we'll need to update the name to reflect that instead so for these places we need a call like:
`super(side + " Agitator");`

aschokking (changes requested):
So everything with a 'side' should get it's name set in constructor super call to include that side

Competition2017/src/competition/subsystems/agitator/AgitatorSubsystem.java
imathno: Doon

aschokking (approved):

PR #59, opened 2017-02-17
Robot To Boiler Command Group

tfrancisco:


Competition2017/src/competition/subsystems/autonomous/MoveToHopperCommandGroup.java
```diff
@@ -0,0 +1,52 @@
...
+        
+        RotateRobotToBoilerCommand rotateRobotToBoiler1 = rotateRobotToBoilerProvider.get();
+        this.addSequential(rotateRobotToBoiler1);
+
+        RotateToHeadingCommand rotateToParallelCommand = rotateToHeadingProvider.get();
+        rotateToParallelCommand.setTargetHeadingProp(angleToParallel);
+        this.addSequential(rotateToParallelCommand);
+
+        DriveForDistanceCommand driveAlongParallel = driveForDistanceProvider.get();
+        driveAlongParallel.setDeltaDistance(distanceFromBoiler.get() * Math.cos(angleToParallel.get()));
```
WasabiFan: This value is only known when the command group starts running. Save the `DriveForDistanceCommand` instance to a private variable, and then in `initialize` for the command group you can compute the distance value and set it.

PR #66, opened 2017-02-19
changed distance calculations to incorporate gears

tfrancisco:


Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -29,12 +32,15 @@
     private final DoubleProperty fVelProp;
     private final DoubleProperty leftDriveEncoderTicksProp;
     private final DoubleProperty rightDriveEncoderTicksProp;
-    private final DoubleProperty ticksPerInch;
     private final DoubleProperty startPositionTicks;
     private final DoubleProperty currentDisplacementInchProp;
+    private final DoubleProperty highGearTicksPerInch;
+    private final DoubleProperty lowGearTicksPerInch;
+    
+    private ShiftSubsystem shift;
```
JohnGilb, line 40: where possible, we've been using longer names for subsystem variables, like `shiftSubsystem`.

JohnGilb (changes requested):
This looks like it is missing the accumulation logic - shouldn't that be in this PR?

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -49,7 +55,8 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager) {
         leftDriveEncoderTicksProp = propManager.createEphemeralProperty("Left drive encoder ticks", 0);
         rightDriveEncoderTicksProp = propManager.createEphemeralProperty("Right drive encoder ticks", 0);
         
-        ticksPerInch = propManager.createPersistentProperty("Ticks per inch", 25.33);
+        highGearTicksPerInch = propManager.createPersistentProperty("Ticks per inch for high gear", 48.8);
+        lowGearTicksPerInch = propManager.createPersistentProperty("Ticks per inch for low gear", 48.8);
```
aschokking, line 59: might be handy to start these at 2 different numbers by default so it'll be more evident they shouldn't match?

PR #69, opened 2017-02-21
Added influx to the robot for visualize data

imathno:
Not ready to be merged going to switch over to another library for efficiency graphing the data.


Competition2017/src/competition/InfluxDBWriter.java
```diff
@@ -0,0 +1,93 @@
...
+public class InfluxDBWriter {
+    private static Logger log = Logger.getLogger(InfluxDBWriter.class);
+   
+    private Configuration config;
+    private DataWriter writer;
+    private boolean dbOnline = false;
+   
+    @Inject
+    public InfluxDBWriter() {
+        config = new Configuration("10.4.88.100", "8086", "", "", "XbotRobotData");
```
WasabiFan, line 29: What is this IP address?

imathno: The RaspberryPI

WasabiFan (changes requested):
While the ability to get this data is great, I don't think it is reasonable to build this into our robot program.

- It requires _another JAR_ to be loaded on startup. Our robot program already takes 5-10 seconds to load from initial invocation, and that time is both detrimental to iteration and devastating if our program has to restart mid-match.
- It depends on an external RasPi to function, which reduces its utility, causes conflicts with other systems and adds more hardware that we need to run a robot.
- It attempts a slow web request when the program loads, and does that for every instance of the writer class.
- It only works if the remote RasPi is on and available when the robot program loads, and requires that the Pi has a static address (which doesn't work at competition).
- It puts more load on the RIO processor with every call to the logging functions, which slows our loop time.
- It increases the chances of code failures bricking our robot.

If you can find a way to make this optional at compilation time, it would be better; but even then, we should be looking for a less intrusive solution instead.

aschokking:
@WasabiFan:
The current library being used is definitely unworkable in terms of way too many web requests, there's another one with better batching support being looked into. 

Agree we need a way to disable this logging completely during competition but the jar itself is pretty tiny and isn't impacting robot startup by anything perceptible.

And yes agreed the static ip thing won't work on the competition routers, open issue there! Would like to touch base on how we've gotten things to settle with the jetson communication under those circumstances.

PR #70, opened 2017-02-22
Influx Database Implementation 

imathno:
We now use a batching method so that sending the data would be more efficient than sending multiple packets a second.


Competition2017/.classpath
```diff
@@ -12,5 +12,11 @@
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib.jar" sourcepath="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="lib" path="lib/influxdb-java-2.5.jar" sourcepath="/home/alex/frc/repos/influxdb-java"/>
+	<classpathentry kind="lib" path="lib/guava-21.0.jar"/>
+	<classpathentry kind="var" path="wpilib" sourcepath="wpilib.sources"/>
```
JohnGilb: Remove this line, and the 3 below it.

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -56,4 +64,17 @@ public void intakeUsingSpeed() {
     public void stop(){
         setPower(0);
     }
+    
+    @Override
+    public void updatePeriodicData() {
+        super.updatePeriodicData();
+        
+        Point primaryMotorPoint = Point.measurement(this.getClass().getSimpleName())
```
JohnGilb: You do this so often, since many of our subsystems have motors. Maybe the XCANTalon should expose a `getTelemetryPoint(String callerName, String tag)` or something like that?

JohnGilb (changes requested):
See comments

JohnGilb (approved):

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -56,4 +64,17 @@ public void intakeUsingSpeed() {
...
+                .time(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
+                .tag("side", side.toString().toLowerCase())
+                .addField("power", masterMotor.get())
+                .addField("current", masterMotor.getOutputCurrent())
```
aschokking: we should get the current velocity as well

Competition2017/src/competition/subsystems/shooter_wheel/ShooterWheelSubsystem.java
```diff
@@ -65,5 +73,18 @@ public void setTargetSpeedForRange(double rangeInInches) {
...
+    
+    @Override
+    public void updatePeriodicData() {
+        super.updatePeriodicData();
+        
+        Point primaryMotorPoint = Point.measurement(this.getClass().getSimpleName())
+                .time(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
+                .tag("side", side.toString().toLowerCase())
+                .addField("power", masterMotor.get())
+                .addField("current", masterMotor.getOutputCurrent())
```
aschokking: we should get the current velocity here too

aschokking (changes requested):

aschokking:
Is there a seriouslycommonlib change for this new method on XCANTalons?

Competition2017/.classpath
```diff
@@ -12,5 +12,7 @@
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="lib" path="/SeriouslyCommonLib/lib/WPILib.jar" sourcepath="/SeriouslyCommonLib/lib/WPILib-sources.jar"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="lib" path="lib/influxdb-java-2.5.jar" sourcepath="/home/alex/frc/repos/influxdb-java"/>
```
WasabiFan: Fix this path

Competition2017/.classpath
```diff
@@ -12,5 +12,7 @@
...
+	<classpathentry kind="lib" path="lib/guava-21.0.jar"/>
```
WasabiFan: Why do we need guava? This should not be a dependency.

Competition2017/robot-program-build.properties
```diff
@@ -23,12 +23,13 @@ guice_assisted.jar=${commonlib.libs}/guice-3.0/guice-assistedinject-3.0.jar
 inject.jar=${commonlib.libs}/guice-3.0/javax.inject.jar
 log4j.jar=${commonlib.libs}/log4j/log4j-1.2.17.jar
 commonsio.jar=${commonlib.libs}/commons-io-2.4/commons-io-2.4.jar
-derby.jar=${commonlib.libs}/derby/derby.jar
+derby.jar=${commonlib.libs}/Derby/derby.jar
 junit.jars=${commonlib.libs}/hamcrest-core-1.3.jar:${commonlib.libs}/junit-4.12.jar
 opencv.jar=${commonlib.libs}/opencv.jar
 ctre.jar=${commonlib.libs}/CTRLib.jar
+influxdb.jars=lib/influxdb-java-2.5.jar:lib/assertj-core-2.6.0.jar:lib/byte-buddy-1.6.5.jar:lib/byte-buddy-agent-1.6.5.jar:lib/converter-moshi-2.1.0.jar:lib/guava-20.0.jar:lib/hamcrest-core-1.3.jar:lib/junit-4.12.jar:lib/logging-interceptor-3.6.0.jar:lib/mockito-core-2.6.8.jar:lib/moshi-1.2.0.jar:lib/objenesis-2.5.jar:lib/okhttp-3.6.0.jar:lib/okio-1.11.0.jar:lib/retrofit-2.1.0.jar:lib/slf4j-api-1.7.23.jar:lib/slf4j-simple-1.7.23.jar
```
WasabiFan: Why do we need this many dependencies? It's a major red flag that something larger is going wrong.

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -199,5 +206,10 @@ public void updatePeriodicData() {
         
         leftDriveEncoderTicksProp.set(leftDrive.getPosition());
         rightDriveEncoderTicksProp.set(rightDrive.getPosition());
+        
+        Point leftPrimaryPoint = leftDrive.getTelemetryPoint(this.getClass().getSimpleName(), "left", true);
```
WasabiFan: This should only be done if the telemetry system is actually in use (here and everywhere else).

Competition2017/src/telemetry/InfluxDBConnection.java
```diff
@@ -0,0 +1,48 @@
...
+    protected final BooleanProperty isEnabledProp;
+    protected final StringProperty influxAddressProp;
+    
+    protected final InfluxDB influxDB;
+    protected final String dbName = "XbotRobotData";
+    
+    @Inject
+    public InfluxDBConnection(XPropertyManager propMan) {
+        isEnabledProp = propMan.createPersistentProperty("InfluxDB logging enabled", false);
+        influxAddressProp = propMan.createPersistentProperty("Influx address", "http://10.4.88.100:8086");
```
WasabiFan: Again, there's no reason that a live telemetry system should depend on an external device.

WasabiFan (changes requested):

Competition2017/.classpath
aschokking: good call, we shouldn't include the sourcepath in here at all, just needed it for some initial debugging

Competition2017/robot-program-build.properties
aschokking: These are the dependencies of this library: https://github.com/influxdata/influxdb-java/blob/master/pom.xml 

Though we can remove lib/junit-4.12.jar, we already have that from the commonlib

Competition2017/src/telemetry/InfluxDBConnection.java
aschokking: Good question, there are several reasons in favor from my perspective:

- Doesn't use up any of the (limited) system resources on the rio (memory, cpu, disk)
- Doesn't require any additional setup/installing of packages in the roborio

aschokking (approved):

Competition2017/src/telemetry/InfluxDBConnection.java
WasabiFan: Two years ago, I implemented a web server that ran from a robot program process and served a web interface which let you control motors remotely. It had no dependencies on additional JARs (although I did later re-implement some of the communication logic with a more efficient library), required no additional set-up and did not impact load times _at all_. My benchmarking revealed that both CPU and memory were immeasurably affected by this system when it was in use, and when no client was connected there was absolutely no impact on the operation of the rest of the processor – a simple in-memory Boolean comparison told you if a client was there to consume data. No external requests were being made and no resources were being consumed. This implementation was a project that I worked on for a couple days in my free time at school, where almost all of that development effort was spent implementing fancy client-side features. We don’t need to speculate over what technical impact the PRed implementation has or what advantages it holds over other solutions, and we have no reason to attempt to convince ourselves that this is the optimal solution: we know empirically and demonstrably that a system like this can be implemented with ease with a fantastically small footprint and no downsides. So, instead of chasing the “coolest” implementation, I think we should be working towards an efficient, simple and robust solution. If we can't find an existing one that fits our needs right now, we shouldn't push to compromise – it’s a replacement for a system that we and thousands of other teams have used for years, and it continues to function well to this day. After all, implementing a browser-based solution for this very problem was something I advocated for in our summer planning and it was shot down within the hour; it can't be that important.

aschokking: It was indeed a very cool project you did and I agree we could roll something ourselves here if we wanted to. That said, this relatively off-the-shelf solution seems to be promising and something being fun/cool is a reasonable reason to pursue something on xbot from (my perspective). 

I think we disagree about the merits of the current solution compared to what would be involved with rolling our own at this point in time (and that's okay). If you're interested in working on a solution here either now or in the summer I would certainly support it (and I'm sorry I don't remember the summer proposal you were referring to before that was shut down).

aschokking (approved):

aschokking:
Would be nice to test this branch out on the summer bot or something before we merge it in (with both the pi plugged in and no plugged in).

WasabiFan (changes requested):
Disregarding the philosophical and design issues with this implementation, it is entirely infeasible for use on our robot. It requires network connectivity which is unavailable, space which is used by essential components, and power which cannot be supplied. It brings with it inherent risk to our robot's functionality and unnecessarily jeopardizes the work of our team as a whole. Its usage will inhibit progress on critical systems and frustrate efforts by other team members. Despite clear and reasonable explanation of these concerns, the issues have not been addressed and have been seemingly ignored, without consulting those who are impacted. This should not and cannot be allowed to continue without fundamental changes to this design, for which we are out of time. If you'd like to consider progressing with this workstream, raise it in the offseason or spend your own time to develop a non-disruptive solution without sidelining team members. I cannot in any reasonable way justify allowing this to be merged in its current state, and you can be assured that any attempts to continue with this past merging these changes will be met with equal pushback from both others and the technicalities of operating a robot.

JohnGilb:
Kaelin, there's some context here you may not be aware of. All of this telemetry is meant to be used to help us tune/investigate the practice robot. This framework makes it relatively simple to chart a number of variables side-by-side to help that effort.

As such, there's no integration challenge for the Competition robot, since it will not be running on the competition robot.

WasabiFan:
The concerns irrespective of whether it's the competition or practice bot. The important factors in relation to that thinking are:

- The code added from this PR is still on-robot in competitions
- The practice robot has the same technical integration challenges as the competition bot

The fact of the matter is, there's no physical room to mount the Pi securely, no power source to supply for its operation, and no Ethernet port to give it connectivity. That doesn't change between the two robots.

aschokking:
Good questions! I appreciate your passion and insights here Kaelin.

- RE code on the competition bot, the code as written uses a boolean property to disable any attempts at network connectivity. I'd be happy to debate pros/cons of that approach vs a heavier handed compile time switch but we have 2.5 weeks before our first competition to land on something there.
- The pi can be powered directly from the roborio's usb port (which I haven't heard of us having any other use cases for to date)
- RE networking, I agree on the new radios we're port constrained if we want to run the jetson and the telemetry system simultaneously. I think in the short-run our shooting tuning won't require vision at the same time. If we find ourselves in the position of needing both we can also run the influxdb server on the drive station laptop with a quick change to the ip address property.

aschokking:
Also upon further reflection, it isn't necessary to have ethernet connected outside of competitions as the pi has onboard wifi and could join the robot's network that way. It's just simpler to do ethernet if available.

WasabiFan:
> RE code on the competition bot, the code as written uses a boolean property to disable any attempts at network connectivity. I'd be happy to debate pros/cons of that approach vs a heavier handed compile time switch but we have 2.5 weeks before our first competition to land on something there.

It's good that you have put the networking code behind a flag, but removing networking attempts does not solve the problem. For a seldom-used feature, this PR roughly doubles the amount of JAR baggage that must be deployed and loaded and adds creation of various objects and invocation of methods every tick that is _absolutely useless_ outside of the short times that this system is used. If that code throws an exception, it's fatal to our robot and our team, for no added value.

The notion that we're pressed for time is erroneous; we already have a property system that can give us data and graph it, and we have easy methods of doing more detailed analysis by enabling a Boolean property on-the-fly without any setup. As you said, your system isn't used during a competition, so having it by then is irrelevant.

> The pi can be powered directly from the roborio's usb port (which I haven't heard of us having any other use cases for to date)

OK, so we can power it from the USB port. Where does the RasPi go? Where does its wire go? How do we make sure that the Pi stays on and the wire doesn't get caught anywhere? We have a team of 30+ people who have made plans that you are trying to disrupt.

> RE networking, I agree on the new radios we're port constrained if we want to run the jetson and the telemetry system simultaneously. I think in the short-run our shooting tuning won't require vision at the same time. If we find ourselves in the position of needing both we can also run the influxdb server on the drive station laptop with a quick change to the ip address property.

It sounds like what you're telling me is that this is _not_ -- and never will be -- a general-purpose system. Instead, it would be used solely for stationary testing of the shooter motors, and would be disconnected and entirely removed at all other times. You would decide you needed to profile shooter data, set the Boolean property, mount and connect the Pi, restart the code, and then run testing; when you're done (within a matter of minutes), you do the inverse of that. The use case for this system is solely while the programming team is in control of the robot while attempting to fix an issue experienced during testing and _not_ when the tuning of the shooters is done and normal work resumes. Is that correct?

> Also upon further reflection, it isn't necessary to have ethernet connected outside of competitions as the pi has onboard wifi and could join the robot's network that way. It's just simpler to do ethernet if available.

That's not a practical option. To connect via Wi-Fi, you need to connect a display and mouse/keyboard, which doesn't work in our practice space.

---

Assuming the above assertion regarding use case is correct, there's a clear way forward: keep this code in a branch, keep the prematurely-merged common lib stuff separate as well, and only deploy from the influx branch when we are doing benchtop shooter tuning. That eliminates the risks to our robot and team for all but the 10 minutes that we use this system and keeps our codebase and deploy clean; if you want to use this system, you can do it, and if you want to use the existing methods that's also available.

aschokking:
Thanks for the feedback @WasabiFan.

Some more specific follow ups:

> If that code throws an exception, it's fatal to our robot and our team, for no added value.

The creation of the objects could be wrapped in a try/catch.

> The notion that we're pressed for time is erroneous;

I didn't mean with the 2.5 weeks statement to imply we were pressed for time, quite the opposite. I meant that the current state of the code in/out question didn't need to block it's usage in the short-run to solve the problem it was meant to and (to the latter conversations in this thread) is something we could change later before the competition as we learn about the ins/outs.

> we have easy methods of doing more detailed analysis by enabling a Boolean property on-the-fly without any setup.

This task had originally come around because while we had a way of generating more detailed data (using the boolean property) the actual analysis involved a lot of slow, manual steps.

> Where does the RasPi go?

This is a good question. One that has been difficult to answer recently without the robot assembled/wired up. Let me begin by asserting that it is not my intention to disrupt the plans of the team and I would appreciate it if you would assume a little bit more positive intent on the behalf of your team mates. Since as planned the rpi wouldn't need to be on the robot during competitions it doesn't need to be situated somewhere that is otherwise legal (eg, could be outside the legal perimeter. I'm not saying such a location that wouldn't interfere definitely will exist in the final config but I would disagree if you said you were 100% certain it could not. 

> It sounds like what you're telling me is that this is not -- and never will be -- a general-purpose system. 
My position is that this is _primarily_ about a particular short-run need, but I can imagine a not so distant future where it (influxdb) would be part of a more general-purpose system. 

>  The use case for this system is solely while the programming team is in control of the robot while attempting to fix an issue experienced during testing and not when the tuning of the shooters is done and normal work resumes. Is that correct?

I disagree with the "and not when the tuning of the shooters is done" piece of this, as that's the primary use-case here, but perhaps I misread your statement? For instance this is data Andrew could use while tuning ball flow rates even if the programming team isn't 'in control' of the robot.

> That's not a practical option. To connect via Wi-Fi, you need to connect a display and mouse/keyboard, which doesn't work in our practice space.

We are not in agreement on this point. With the rpi as configured we can `ssh` into it via ethernet and set it up to join a wifi network that way (which is how I generally need to set them up). Another easy option here is to setup the pi to join a hotspot wifi network if it sees it first and ssh that way. 

> keep this code in a branch, keep the prematurely-merged common lib stuff separate as well, and only deploy from the influx branch when we are doing benchtop shooter tuning.

I think this could be a valid option in the short-run for the shooting tuning testing. 

-----

Stepping back, I think we need to separate out the rpi from influx here as a lot (not all) of the current discussion is centering around the rpi but the influxdb server can be run on the driver station laptop as well. We just hadn't selected a laptop at the time this initiative started and the convenience of having data setup such that any laptop could access it was appealing. That said, we've now selected and stood up the dedicated drive laptop (hopper) and setting up the server there over using the rpi is probably the better route forward. 

Because I think influx has the potential to be a good general-purpose solution down the line, I would like to include the code in master so we can test/vet that over the coming (now) 2 weeks.

WasabiFan:
> The creation of the objects could be wrapped in a try/catch.

That would be a wise addition, and with each step the fatality of potential problems decreases. However, the more central issues raised alongside remain unaddressed. 

> I meant that the current state of the code in/out question didn't need to block it's usage in the short-run to solve the problem it was meant to and (to the latter conversations in this thread) is something we could change later before the competition as we learn about the ins/outs.

That sounds reasonable. I would like to keep in mind that that any code that enters `master` is almost certainly going to stay there; once a merge is performed, there won't be anyone with the combination of time and drive to repair it. So, merging first and thinking later won't be a solution to the problem.

> This task had originally come around because while we had a way of generating more detailed data (using the boolean property) the actual analysis involved a lot of slow, manual steps.

I'm confident that you'd be able to write a Python script in a dozen lines to simplify the process down to a matter of seconds. All it needs to do is:
- SCP the log/data file from the RIO
- Parse out the relevant lines
- Open Excel and graph the data (likely with the COM API)

That script would take you from captured data to analyzed data extremely quickly, with a miniscule amount of manpower required to get it up-and-running. I agree that a web portal would be nicer in some ways, but for detailed analysis I'd prefer it in Excel anyway, and the overall impact of a simple solution like the one above is almost nothing.

> Let me begin by asserting that it is not my intention to disrupt the plans of the team and I would appreciate it if you would assume a little bit more positive intent on the behalf of your team mates.

I began without any preconceptions; I asked about the IP address lines without judgement, looking to understand the system without jumping to faulty conclusions. The worrying part for me was the fact that there was no clear way forward to find a non-impactful solution to integration issues, and as a result we're discussing compromises rather than enhancements.

> Since as planned the rpi wouldn't need to be on the robot during competitions it doesn't need to be situated somewhere that is otherwise legal (eg, could be outside the legal perimeter. I'm not saying such a location that wouldn't interfere definitely will exist in the final config but I would disagree if you said you were 100% certain it could not.

I am indisputably unsure of whether or how such a system can be mounted, however there is enough which is known to warrant concern. There is no sense in arguing over the possible mounting configurations; with proper consideration and a mind for practicality, attempts to add it to the robot will either succeed with a reasonable conclusion or fail and give useful information. As long as a half-considered solution isn't forced through, I see no harm it would cause.

> I disagree with the "and not when the tuning of the shooters is done" piece of this, as that's the primary use-case here, but perhaps I misread your statement? For instance this is data Andrew could use while tuning ball flow rates even if the programming team isn't 'in control' of the robot.

It sounds like what you're saying is that you expect it to be used more continuously. That calls into question previous conclusions, but most important is that this is incompatible with existing plans if you intend to use an Ethernet port. The Jetson gets first-priority access to those ports as a part of the core system which operates at competition.

---

As a general observation, the issues that have been identified with this system entirely stem from the fact that it relies on an external device which isn't otherwise needed. In line with my questions earlier, I still see no reason that there should be a separate device that a logging system depends on, and the severe disadvantages of such a design weigh heavily on its potential. While the overhead of loading an unused server is negligible, sending packets outward to an intermediary with no use for them is an intense operation for our RIO to handle. Accordingly, I would be much more supportive of a design which were chosen or designed with our environment in mind.

aschokking:
We've been able to get away with the smartdashboard's built in graphs so far, let's take another look at this in the off-season. 

Will pull the jars/code from the SeriouslyCommonLib in the meantime too, would be easier to play with if everything was in 1 repo anyway.

PR #77, opened 2017-02-26
Implement ability to drive on curve.

SethNguyenXBOT:
… curve and tests.


aschokking:
Is this PR still something we want or is it covered by the other 2 branches (yours and Andy's)

JohnGilb:
Closing for now - let's revisit this in the off-season.

PR #82, opened 2017-03-03
Rationalized OI button mapping.

SethNguyenXBOT:


Competition2017/src/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -100,9 +100,9 @@ public void setupShooterBeltCommands(
             ShooterBeltsManagerSubsystem shooterBeltsSubsystem)
     {
         RunShooterBeltPowerCommand runBeltLeft = new RunShooterBeltPowerCommand(shooterBeltsSubsystem.getLeftBelt());
-        oi.leftButtons.getifAvailable(2).whileHeld(runBeltLeft);
+        oi.controller.getXboxButton(XboxButton.LeftStick).whileHeld(runBeltLeft);
```
aschokking: So to fire the shooter someone will hold down on a given side: Stick, Bumper and Trigger?
I'm wondering if we can simplify that down a bit now that we have a mostly functioning robot, for instance can we have 1 button that spins up the shooter and runs the belt while at speed?

JohnGilb (changes requested):
We need the commands that set the shooters to speed goals to be on the gamepad. Right now they are on the joysticks.

WasabiFan:
Remember that you have the D-pad to work with too... that might make more sense for ascend/descend.

aschokking:
unless something has changed, the last time it was tested I believe the dpad wasn't functioning

WasabiFan:
How so? The interface was implemented, so if there's a bug preventing it from getting data, that should be investigated.

JohnGilb (approved):

aschokking (approved):

aschokking:
we should write this mapping down somewhere that we can keep near the OI

PR #88, opened 2017-03-04
Setup for autonomous program that breaks the baseline for 5 points

tfrancisco:


Competition2017/src/competition/subsystems/autonomous/selection/SetupBreakBaselineCommand.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public final DriveForDistanceCommand breakBaselineAuto;
+    // distance to break baseline from a position with both back wheels on the wall
+    private DoubleProperty distanceFromWallToBaseline;
+
+    public SetupBreakBaselineCommand(AutonomousCommandSelector autonomousCommandSelector,
+            DriveForDistanceCommand driveForDistance, XPropertyManager propMan) {
+        super(autonomousCommandSelector);
+        distanceFromWallToBaseline = propMan.createEphemeralProperty("distance from wall to break baseline in inches", 96);
+        this.breakBaselineAuto = driveForDistance;
```
aschokking: my only thought here is that the name of the auto command will be just "DriveForDistanceCommand" but if we only have the 1 auto program using that it might not be so bad

aschokking (approved):

Competition2017/src/competition/subsystems/autonomous/selection/SetupBreakBaselineCommand.java
```diff
@@ -0,0 +1,26 @@
...
+public class SetupBreakBaselineCommand extends BaseAutonomousCommandSetter {
+
+    public final DriveForDistanceCommand breakBaselineAuto;
+    // distance to break baseline from a position with both back wheels on the wall
+    private DoubleProperty distanceFromWallToBaseline;
+
+    public SetupBreakBaselineCommand(AutonomousCommandSelector autonomousCommandSelector,
+            DriveForDistanceCommand driveForDistance, XPropertyManager propMan) {
+        super(autonomousCommandSelector);
+        distanceFromWallToBaseline = propMan.createEphemeralProperty("distance from wall to break baseline in inches", 96);
```
aschokking: This should be a persistent property not ephemeral, otherwise it'll forget the value we set every time we reboot the robot.

aschokking (dismissed):

JohnGilb (approved):

PR #89, opened 2017-03-05
Add vision server and drive logic

WasabiFan:
This adds the vision server and drive logic. The changes that will immediately affect the robot at runtime are:
- A server is started on UDP port 5800. This code is surrounded in error handling so it shouldn't be able to cause fatal errors, and starting a server requires negligible resources; a listener is registered and a new thread sleeps waiting for the `receive()` call to return.
- Four testing commands are added to left joystick buttons 7, 8, 9 and 10.

The latter half of this diff adds the source of a JSON parsing library (so that no additional JARs have to be loaded).


Competition2017/src/competition/networking/NetworkedCommunicationServer.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+        
+        public void setNewPacketHandler(Consumer<OffboardCommPacket> handler) {
+            this.packetHandler = handler;
+        }
+
+        public void stopServer() {
+            this.isRunning = false;
+            serverSocket.close();
```
imathno: Should we log when the server is stopping?

Competition2017/src/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -170,6 +173,24 @@ public void setupDriveCommand(
     }
     
     @Inject
+    public void setupVisionCommands(
+            OperatorInterface oi,
+            RotateRobotToBoilerCommand rotateCommand,
+            DriveToBoilerWithVisionCommandGroup driveToBoilerCommand,
+            SetRobotHeadingCommand setHeadingCommand,
+            RotateToHeadingCommand rotateToHeadingCommand
+    )   {
```
JohnGilb: At this point, we've moved all debug functions to the dashboard instead of the joystick, meaning that the upper joystick buttons (the easier ones to press) are available. I suggest putting your commands up there now, to increase the chance that the drivers will try them out and acclimate to them.

Competition2017/src/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -170,6 +173,24 @@ public void setupDriveCommand(
...
+        oi.leftButtons.getifAvailable(7).whileHeld(rotateCommand);
+        oi.leftButtons.getifAvailable(8).whileHeld(driveToBoilerCommand);
+        
+        setHeadingCommand.setHeadingToApply(248);
```
JohnGilb: Is there some significance to this number?

JohnGilb: I see it's "Angle of line parallel to boiler". I'd suggest putting that as a static variable in the pose subsystem or the vision subsystem.

Competition2017/src/competition/subsystems/drive/commands/CalculateDistanceOfBoilerParallelCommand.java
```diff
@@ -0,0 +1,71 @@
...
+    public void setParallelAngle(double angle) {
+        parallelAngleSupplier = () -> angle;
+    }
+    
+    public void setParallelAngleProp(DoubleProperty prop) {
+        parallelAngleSupplier = () -> prop.get();
+    }
+
+    @Override
+    public void initialize() {
```
JohnGilb: Most of our commands include an "initializing" log to help us track runtime flow. I'd recommend it here and in other new commands.

Competition2017/src/competition/subsystems/drive/commands/DriveInfinitelyCommand.java
```diff
@@ -0,0 +1,38 @@
...
+        powerTargetSupplier = () -> power;
+    }
+    
+    public void setDrivePowerProp(DoubleProperty prop) {
+        powerTargetSupplier = () -> prop.get();
+    }
+
+    @Override
+    public void initialize() {
+        power = powerTargetSupplier == null ? 0 : powerTargetSupplier.getAsDouble();
```
JohnGilb: same "Initializing"

Competition2017/src/competition/subsystems/vision/VisionSubsystem.java
```diff
@@ -0,0 +1,165 @@
...
+    private volatile int numPacketsSinceReset = 0;
+
+    private volatile boolean isConnected = false;
+    
+    private DoubleProperty connectionTimeoutThreshold;
+    private DoubleProperty connectionReportInterval;
+
+    @Inject
+    public VisionSubsystem(WPIFactory factory, XPropertyManager propManager, OffboardCommunicationServer server) {
+        log.info("Creating VisionSubsystem");
```
JohnGilb: with the new logging settings Alex did, you don't need to include the name, so this can just be `log.info("Creating");`

Competition2017/src/competition/subsystems/vision/commands/RotateRobotToBoilerCommand.java
```diff
@@ -0,0 +1,58 @@
...
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        DetectedBoiler target = visionSubsystem.getTrackedBoiler();
+        
+        rotationPid.setIMask(target == null);
+        double power = rotationPid.calculate(0, target == null ? 0 : target.offsetX);
```
JohnGilb: This is a small thing, but I've taken to doing `rotationPid.calculate(target == null ? 0 : target.offsetX, 0);` . This means that if you need to turn left, the PID will give you a positive result, which meshes nicely with the global "positive rotation is left rotation"

JohnGilb (approved):

WasabiFan:
I've updated this PR after review; this code has had a few hours of use on-robot (I was working with it until after midnight on Tuesday 😁 ) and hasn't caused any problems since I fixed that synchronization issue. The _drive to boiler with heuristics_ code is still broken, but that's an issue with the underlying drive command.

Can we aim to merge this today? I'd like to know that the base vision code is integrated so we can give it some time to soak; I will probably branch and work on some additional user-friendly optimizations now.

JohnGilb:
Overall still looks good - you can now take advantage of red/blue alliance detection via the pose subsystem , since you'll need two "parallel to boiler" angles (I think?)

PR #90, opened 2017-03-06
Use new field pose 

WasabiFan:
https://github.com/Team488/SeriouslyCommonLib/pull/83


aschokking (approved):

WasabiFan:
I tested this on Tuesday and confirmed that the heuristics-based drive behavior with this math is the same (read: just as wrong :wink:) as it was with the old.

JohnGilb:
Closing for now - let's revisit this in the off-season.

PR #96, opened 2017-03-11
Added reverse to agitator using the left and right stick

imathno:
To reverse left press down on the left stick and same for the right side.


JohnGilb (changes requested):
You need to remove the existing commands that use the sticks.

JohnGilb (approved):

PR #97, opened 2017-03-12
Add ability to ignore limit switches when climbing

JohnGilb:
Adds an `AscendDangerouslyCommand` that will run the climber even if limit switches are pressed.

### DIFF
--- a/Competition2017/src/competition/operator_interface/OperatorCommandMap.java
+++ b/Competition2017/src/competition/operator_interface/OperatorCommandMap.java
@@ -58,6 +58,7 @@ public class OperatorCommandMap {
             RopeAlignerCommand aligner)   
     {
         oi.controller.getXboxButton(XboxButton.X).whileHeld(ascend);
+        oi.operatorPanelButtons.getifAvailable(4).whileHeld(ascend);
         
         oi.controller.getXboxButton(XboxButton.Start).whileHeld(aligner);
         oi.controller.getXboxButton(XboxButton.Back).whileHeld(ascendDangerously);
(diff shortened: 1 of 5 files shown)


Competition2017/src/competition/operator_interface/OperatorCommandMap.java
```diff
@@ -52,14 +52,15 @@ public void setupMyCommands(
...
+            DescendClimbingCommand descend,
+            AscendSafelyCommand ascend,
+            AscendDangerouslyCommand ascendDangerously,
             RopeAlignerCommand aligner)   
     {
         oi.controller.getXboxButton(XboxButton.X).whileHeld(ascend);
         
         oi.controller.getXboxButton(XboxButton.Start).whileHeld(aligner);
-        
-        oi.operatorPanelButtons.getifAvailable(4).whileHeld(ascend);
```
WasabiFan: Replacing a panel button with an Xbox button? Are we no longer using the panel button?

Competition2017/src/competition/subsystems/climbing/ClimbingSubsystem.java
```diff
@@ -56,6 +58,14 @@ public void descend() {
     public void setBrake(boolean brakeOn) {
         brakeSolenoid.set(brakeOn);
     }
+    
+    public boolean isLimitSwitchPressed() {
+        return climbingMotor.isForwardLimitSwitchClosed();
+    }
+    
+    public void setEnableSafeties(boolean isEnabled) {
+        climbingMotor.enableLimitSwitches(isEnabled, isEnabled);
```
WasabiFan: Earlier, the reverse limit switch is configured to be disabled universally; I think we want to do that here as well.

WasabiFan (changes requested):
We should make sure to test this and confirm that the Talons process the limit switch enable/disable in such a way that they take effect immediately.

Competition2017/src/competition/subsystems/climbing/ClimbingSubsystem.java
JohnGilb: Yes, we should be consistent.

Competition2017/src/competition/operator_interface/OperatorCommandMap.java
JohnGilb: Ah, didn't mean to also replace the panel. Thanks for noticing.

PR #99, opened 2017-03-12
Feature/trim shooters

JohnGilb:
This changes the two momentary switches on the operator panel to serve as "trim" for the left/right shooter wheel speeds.


WasabiFan:
Without any indication of the current value, limits on the trimmed result, or other feedback and protection mechanisms (more than one of which would be required), I think this would put us at more risk than it's worth. If the button gets depressed accidentally or a software/hardware glitch begins firing button events erroneously, we lose control of our shooters... and the changes that are made will persist even through resets.

I think this would be better if we put a physical slider on the OI with requisite override controls, or potentially we could take advantage of Xbox controller vibration to pulse the controller when trim buttons are pressed. We should also store the trimmed speed separately from the persistent property so that it is cleared when robot code is reloaded.

JohnGilb:
Changes made to add some protection and make the trim ephemeral.

Overall, though this adds risk, what we saw during drive practice is that the drive team often needs immediate access to live shooter tuning, or we stand to miss a lot of shots.

PR #102, opened 2017-03-15
Failing heuristics-based drive test

WasabiFan:
This is my attempt to repro the bad behavior I was seeing earlier. It doesn't seem to be exactly the same thing, but as far as I can tell this test should still be passing. It might be worth trying with a heading of around `-30` degrees, but I don't have time to try that out now.


JohnGilb:
Closing for now - we'll revisit advanced motion planning in the offseason =]

PR #109, opened 2017-03-18
Add pair of LIDARs to pose subsystem

WasabiFan:
I figure that this needs to be tested on-robot to make sure it doesn't cause any crashes.


JohnGilb (approved):
I'll have somebody add some double properties to track distance on the dashboard

PR #113, opened 2017-03-22
Add unobtrusive telemetry system

WasabiFan:
This is an implementation that I originally wrote a few weeks back while waiting for vision access. The idea is that you can flip a Boolean property to start telemetry collection and the robot code will stream `DoubleProperty`s to a file. You can then run a basic Python script which dumps the data into Excel and plots it. This is super simple yet has all the power of Excel, and consumes no resources (hardware or software) unless activated.


WasabiFan:
@JohnGilb I'm seeing a behavior where the telemetry system is constantly being uninitialized and reinitialized while it is enabled (via the dashboard property); could you take a look at this and see if you can spot the issue? The log lines in `DiskDeferredTelemetryLogger` are the ones that are being run obnoxiously.

Competition2017/src/competition/subsystems/BaseXCANTalonSpeedControlledSubsystem.java
```diff
@@ -151,12 +164,13 @@ public void updatePeriodicData() {
...
-                 + systemOutputPower.get() + "," 
-                 + systemTalonError.get() + ","
-                 + systemCurrentSpeed.get());
+            if(!telemetryLogger.isInitialized()) {
+                telemetryLogger.initialize(name, systemOutputPower, systemTalonError, systemCurrentSpeed);
+            }
+            telemetryLogger.collectNewDataset();
+        }
+        else if(telemetryLogger.isInitialized()) {
+            telemetryLogger.uninitialize();
```
JohnGilb, line 173: I've looked at this for quite a while and can't see anything wrong. The only thing that would make sense is if you are hitting some sort of exception when creating the `out` object, which would cause you to keep initializing over and over.

But then, you wouldn't see the log call of "uninitialize", so that doesn't seem reasonable.

WasabiFan: And I would expect to see the error log, which I'm not seeing 🤔

JohnGilb (approved):
Generally I approve, though we definitely don't want infinite initialize loops. =P

WasabiFan:
I figured out that disabling the periodic data updates for all other `BaseXCANTalonSpeedControlledSubsystem`s and their telemetry loggers resolved the issue; I can't say why the additional instances cause an issue. The rest of the system works and I was able to plot the data in Excel without delay. @aschokking Do you see anything wrong with the logic spanning multiple systems?

Competition2017/src/competition/DiskDeferredTelemetryLogger.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.log4j.Logger;
+
+import edu.wpi.first.wpilibj.Timer;
+import xbot.common.properties.DoubleProperty;
+
+public class DiskDeferredTelemetryLogger implements DeferredTelemetryLogger {
+    private static String baseTelemetryFileName = "/home/lvuser/488-telemetry/";
```
WasabiFan, line 20: note to self: final

WasabiFan (commented):
I think this is an issue relating to duplicate or erroneously re-used instances of subsystems or the logger class; it might be the instances themselves or it might be the name fields of the various objects, which are used to create properties on the dashboard. I'm thinking that every subsystem is accessing the same instance of the logger, and since most are disabled yet there is one which I have enabled, they're fighting over that single instance to initialize and uninitialize it until it matches the setting for that particular instance. I don't see a reason that the injector would treat the telemetry logger as a singleton though.

My plan is to add log lines which specify that an instance of the telemetry class is being constructed and include the memory instance info so we can tell them apart... That should make it clear whether there is, in fact, only one instance being created.

imathno (approved):

WasabiFan:
This is confirmed working on-robot.

aschokking:
This one is probably reasonable to merge in early today to add the extra logging functionality and give it time to be proven before real matches start.

PR #117, opened 2017-03-24
Update shooter inversions from practice bot

WasabiFan:
These are the inversion changes made today. I'm not sure if we want to integrate these or not.


aschokking:
So these are the net different inversions vs the competition bot?

JohnGilb (approved):

PR #118, opened 2017-03-24
Move robot to ideal range using Lidar.

JohnGilb:
There's one thing I don't like here - I needed test access to the rangefinders on the PoseSubystem. My initial approach was to create a "CompetitionTestPoseSubsystem" and bind that in the TestModule found in `BaseTest`.

However, there's already a binding defined in `UnitTestModule`, and I couldn't find a graceful way to override it.

It looks like there are some solutions to this problem, but they involve a bit of an overhaul - so in the short term I just made a public method on the PoseSubsystem to get the lidar. =\


Competition2017/src/competition/subsystems/drive/commands/DriveToShootingRangeCommand.java
```diff
@@ -0,0 +1,52 @@
...
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        double power = driveToRangePidManager.calculate(
+                poseSubsystem.getIdealShootingRange(), 
+                poseSubsystem.getLidarFrontDistanceAverage());
+        
+        // However, in this case, having a currentDistance of 10 and a goal distance of 20 means you need to reverse,
```
WasabiFan, line 39: Why not invert the operands for `calculate` instead?

Competition2017/src/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -28,11 +28,9 @@
     private final DoubleProperty headingFacingRedBoiler;
     private final DoubleProperty distanceBetweenDistanceSensorsProp;
     private final DoubleProperty breakBaselineMaxTime;
+    private final DoubleProperty idealShootingRange;
     
-    private final DoubleProperty leftLidarDistance;
```
WasabiFan: Why were these removed? They're there for debug readouts.

Competition2017/src/competition/subsystems/pose/PoseSubsystem.java
```diff
@@ -103,10 +99,16 @@ public double getBreakBaselineMaximumTime() {
         return breakBaselineMaxTime.get();
     }
     
+    public double getIdealShootingRange() {
+        return idealShootingRange.get();
+    }
+    
+    public DistanceSensorPair getLidar() {
```
WasabiFan, line 109: Hmmm... I see what you mean. I can't think of a better solution, so I'd say we should go with this for now.

Competition2017/tests/competition/subsystems/pose/TestCompetitionPoseSubsystem.java
```diff
@@ -0,0 +1,23 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+import competition.subsystems.drive.DriveSubsystem;
+import xbot.common.controls.sensors.DistanceSensorPair;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class TestCompetitionPoseSubsystem extends PoseSubsystem {
```
WasabiFan: Is this used? I think you decided to go with the other approach.

WasabiFan (approved):
Looks good :+1:; I think a file was accidentally included in the PR though.

Competition2017/src/competition/subsystems/drive/commands/DriveToShootingRangeCommand.java
JohnGilb: I've gone back and forth on this a few times, and I'm kind of leaning towards splitting out "inversion" when we do PID. I want people reading through the code (especially students) to see a constant use of goal/current, given the `calculate` function signature, and then see a deliberate inversion in a following line.

JohnGilb: That being said, there's already lots of examples of the other way around (some of which written by me), so I'm not too married to the concept.

Competition2017/tests/competition/subsystems/pose/TestCompetitionPoseSubsystem.java
JohnGilb: Hah, gotta remove this =]

Competition2017/src/competition/subsystems/pose/PoseSubsystem.java
JohnGilb: Odd... wonder how this happened.

PR #119, opened 2017-03-25
 Fire tracer rounds

SethNguyenXBOT:


Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -53,6 +57,14 @@ public void intakeUsingSpeed() {
         setTargetSpeed(beltIntakeTargetSpeed.get());
     }
     
+    public void intakeUsingTracerSpeed(){
```
JohnGilb: This should be `intakeUsingTracerPower()`

Competition2017/src/competition/subsystems/BaseXCANTalonSpeedControlledSubsystem.java
```diff
@@ -139,6 +139,10 @@ public boolean isAtSpeed() {
         return Math.abs(getSpeed() - systemTargetSpeed.get()) <= systemSpeedThresh.get();
     }
     
+    public boolean isAtTracerSpeed(){
+        return Math.abs(getSpeed()/2 - systemTargetSpeed.get()/2) <= systemSpeedThresh.get()/2;
```
aschokking: I'm a little confused about this logic, if everything is halved is it any different?

Competition2017/src/competition/subsystems/shooter_belt/commands/FireTracerMode.java
```diff
@@ -0,0 +1,27 @@
+package competition.subsystems.shooter_belt.commands;
+
+import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import xbot.common.command.BaseCommand;
+
+public class FireTracerMode extends BaseCommand {
```
JohnGilb: I would rename this to `RunBeltSlowlyUsingSpeedCommand`, as there's nothing really "tracer-like" about this, since it just operates belt

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -33,8 +35,10 @@ public ShooterBeltSubsystem(
         this.side = side;
         
         intakePowerProperty = propManager.createPersistentProperty("ShooterBelt intake power", 0.5);
+        intakeTracerPowerProperty = propManager.createPersistentProperty("ShooterBelt intake power", 0.25);
```
aschokking: this property has the same name as the one above it so there will only be one on the dashboard

Competition2017/src/competition/subsystems/shooter_belt/commands/FireTracerMode.java
JohnGilb: this should be `extends BaseShooterBeltCommand`

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
aschokking: these two setters are also a little confusing, they both sound like they're setting tracer speeds?

Competition2017/src/competition/subsystems/shooter_belt/commands/FireTracerMode.java
```diff
@@ -0,0 +1,27 @@
...
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        if(shooterBeltSubsystem.isAtSpeed()){
+            shooterBeltSubsystem.intakeUsingSpeed();
```
JohnGilb: this should be `intakeTracerUsingSpeed`

aschokking (changes requested):

Competition2017/src/competition/subsystems/shooter_belt/commands/FireTracerRoundsCommand.java
```diff
@@ -0,0 +1,29 @@
+package competition.subsystems.shooter_belt.commands;
+
+import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelSubsystem;
+
+public class FireTracerRoundsCommand extends BaseShooterBeltCommand {
```
JohnGilb: rename to `RunBeltSlowlyUsingPowerCommand`

Competition2017/src/competition/subsystems/shooter_belt/commands/FireTracerRoundsCommand.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        if(shooterWheelSubsystem.isAtSpeed()){
```
JohnGilb: Replace the entire execute method with `shooterBeltSubsystem.intakeUsingTracerPower();`

Competition2017/src/competition/subsystems/BaseXCANTalonSpeedControlledSubsystem.java
JohnGilb: I think this method is unnecessary - the existing `isAtSpeed()` command should work.

JohnGilb (dismissed):
Many changes needed, especially around naming.

Competition2017/src/competition/subsystems/shoot_fuel/FireLeftTracerRoundsCommandGroup.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.inject.Inject;
+
+import competition.subsystems.agitator.AgitatorsManagerSubsystem;
+import competition.subsystems.agitator.commands.IntakeAgitatorCommand;
+import competition.subsystems.shooter_belt.ShooterBeltsManagerSubsystem;
+import competition.subsystems.shooter_belt.commands.RunBeltSlowlyUsingPowerCommand;
+import competition.subsystems.shooter_wheel.ShooterWheelsManagerSubsystem;
+import edu.wpi.first.wpilibj.command.CommandGroup;
+
+public class FireLeftTracerRoundsCommandGroup extends CommandGroup {
```
WasabiFan: Make this generic so they can work for either side

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        if(shooterBeltSubsystem.isAtSpeed()){
```
WasabiFan: If the _belts_ are at speed?

Competition2017/src/competition/subsystems/shoot_fuel/FireLeftTracerRoundsCommandGroup.java
SethNguyenXBOT: What do you mean by, "Make this generic"?

WasabiFan: Make a single command "FireTracerRoundsCommandGroup" which has a `setSide` method which allows you to select the side that it runs on. That means that we don't have to make changes to both instances of this command (or, more likely, forget to!).

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyCommand.java
WasabiFan: This still seems odd. Why do you need to check if the belts are at speed? Why not the shooter?

Competition2017/src/competition/subsystems/shoot_fuel/FireLeftTracerRoundsCommandGroup.java
WasabiFan: Actually, you might want to take a look at `IntakeAgitatorCommand ` to see how that is done.

Competition2017/src/competition/subsystems/shoot_fuel/FireTracerRoundsCommandGroup.java
```diff
@@ -0,0 +1,37 @@
...
+                        shooterWheelsManagerSubsystem.getShooterWheelWithRobotSide(side));
+        
+        IntakeAgitatorCommand runAgitator = 
+                new IntakeAgitatorCommand(agitatorsManagerSubsystem.getAgiatorWithRobotSide(side));
+        
+        this.addParallel(runTracerPower);
+        this.addParallel(runAgitator);
+    }
+    
+    public void setSide(RobotSide side) {
```
WasabiFan: This can only be called after the constructor, so it seems like `side` will always be `null` at the time that its value is passed into the child commands.

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -53,6 +57,14 @@ public void intakeUsingSpeed() {
         setTargetSpeed(beltIntakeTargetSpeed.get());
     }
     
+    public void intakeUsingTracerPower(){
+        setPower(intakeTracerPowerProperty.get());
+    }
+    
+    public void intakeTracerUsingSpeed(){
```
WasabiFan: Make name formats match

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltsManagerSubsystem.java
```diff
@@ -12,57 +12,64 @@
...
-            rightPIDValues = pidFactory.createPIDPropertyManager("RightBelt", 0, 0, 0, 0);
+    @Inject
+    public ShooterBeltsManagerSubsystem(WPIFactory factory, XPropertyManager propManager, PIDFactory pidFactory) {
+        log.info("Creating");
+        leftPIDValues = pidFactory.createPIDPropertyManager("LeftBelt", 0, 0, 0, 0);
+        rightPIDValues = pidFactory.createPIDPropertyManager("RightBelt", 0, 0, 0, 0);
             
-            createLeftAndRightBelts(factory, propManager);
-        }
+        createLeftAndRightBelts(factory, propManager);
```
WasabiFan: Remove trailing space

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltsManagerSubsystem.java
```diff
@@ -12,57 +12,64 @@
...
-                    propManager);
-        }
+        rightBelt = new ShooterBeltSubsystem(
+                RobotSide.Right,
+                rightMotorIndex,
+                invertRight,
+                invertRightSensor,
+                factory, 
+                rightPIDValues,
+                propManager);
```
WasabiFan: Remove trailing space

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -33,8 +35,10 @@ public ShooterBeltSubsystem(
         this.side = side;
         
         intakePowerProperty = propManager.createPersistentProperty("ShooterBelt intake power", 0.5);
+        intakeTracerPowerProperty = propManager.createPersistentProperty("ShooterBelt intake tracer power", 0.25);
```
WasabiFan: It's unclear what these are. Are the two properties supposed to be equivalent values for running the belt in either speed or power mode?

Competition2017/src/competition/subsystems/shooter_belt/commands/BaseShooterBeltCommand.java
```diff
@@ -1,14 +1,23 @@
 package competition.subsystems.shooter_belt.commands;
 
 import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelSubsystem;
 import xbot.common.command.BaseCommand;
 
 public abstract class BaseShooterBeltCommand extends BaseCommand {
     
     protected final ShooterBeltSubsystem shooterBeltSubsystem;
+    protected final ShooterWheelSubsystem shooterWheelSubsystem;
```
WasabiFan: What do you think about making a `BaseChooterBeltAndWheelCommand` which has this in it instead of a nullable field here?

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyCommand.java
```diff
@@ -0,0 +1,26 @@
+package competition.subsystems.shooter_belt.commands;
+
+import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelSubsystem;
+
+public class RunBeltSlowlyCommand extends BaseShooterBeltCommand {
```
WasabiFan: Can we use the word "tracer" in the name and point out that it only runs when the wheel is at speed?

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/shooter_belt/commands/BaseShooterBeltCommand.java
imathno: Added

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltSubsystem.java
```diff
@@ -35,7 +36,7 @@ public ShooterBeltSubsystem(
         this.side = side;
         
         intakePowerProperty = propManager.createPersistentProperty("ShooterBelt intake power", 0.5);
-        intakeTracerPowerProperty = propManager.createPersistentProperty("ShooterBelt intake tracer power", 0.25);
+        intakeTracerPowerProperty = propManager.createPersistentProperty("ShooterBelt intake tracer power", 0.5);
```
WasabiFan: This is the same as the regular power

Competition2017/src/competition/subsystems/shooter_belt/ShooterBeltsManagerSubsystem.java
```diff
@@ -12,57 +12,64 @@
...
+        return leftBelt;
+    }
 
-        public ShooterBeltSubsystem getRightBelt() {
-            return rightBelt;
+    public ShooterBeltSubsystem getRightBelt() {
+        return rightBelt;
+    }
+        
+    public ShooterBeltSubsystem getBeltWithRobotSide(RobotSide side) {
```
WasabiFan: `with`->`for`?

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyUsingPowerCommand.java
```diff
@@ -0,0 +1,26 @@
+package competition.subsystems.shooter_belt.commands;
+
+import competition.subsystems.hybrid.commands.BaseShooterBeltAndWheelCommand;
+import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelSubsystem;
+
+public class RunBeltSlowlyUsingPowerCommand extends BaseShooterBeltAndWheelCommand {
```
WasabiFan: This seems to be a dupe of `RunBeltTracerModeCommand`

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyUsingPowerCommand.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        if (shooterWheelSubsystem.isWheelAtSpeed()) {
```
WasabiFan: What's the difference between `isWheelAtSpeed` and `isAtSpeed`?

WasabiFan (approved):
We'll need to test this on-robot

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltSlowlyUsingPowerCommand.java
imathno: Has logic that prevents weird things from happening

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltTracerPowerMode.java
```diff
@@ -0,0 +1,26 @@
+package competition.subsystems.shooter_belt.commands;
+
+import competition.subsystems.hybrid.commands.BaseShooterBeltAndWheelCommand;
+import competition.subsystems.shooter_belt.ShooterBeltSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelSubsystem;
+
+public class RunBeltTracerPowerMode extends BaseShooterBeltAndWheelCommand {
```
aschokking: I think there might be a problem here, this base class requires the shooterWheel which means the wheel wouldn't actually spin to shoot because this command only actuates the belts?

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltTracerSpeedModeCommand.java
```diff
@@ -0,0 +1,27 @@
...
+    
+    @Override
+    public void initialize() {
+        log.info("Initializing");
+    }
+
+    @Override
+    public void execute() {
+        if(shooterWheelSubsystem.isAtSpeed()){
+            shooterBeltSubsystem.intakeUsingTracerSpeed();
```
aschokking, line 23: I would advocate for using Power instead of Speed in the short-run since speed isn't currently tuned/verified on the comp bot and if power works that's much simpler to verify this concept.

Competition2017/src/competition/subsystems/hybrid/commands/BaseShooterBeltAndWheelCommand.java
```diff
@@ -0,0 +1,18 @@
...
+public abstract class BaseShooterBeltAndWheelCommand extends BaseCommand {
+    
+    protected final ShooterBeltSubsystem shooterBeltSubsystem;
+    protected final ShooterWheelSubsystem shooterWheelSubsystem;
+    
+    public BaseShooterBeltAndWheelCommand(ShooterBeltSubsystem shooterBeltSubsystem, ShooterWheelSubsystem shooterWheelSubsystem) {
+        this.shooterBeltSubsystem = shooterBeltSubsystem;
+        this.shooterWheelSubsystem = shooterWheelSubsystem;
+        this.requires(shooterBeltSubsystem);
+        this.requires(shooterWheelSubsystem);
```
aschokking: Yeah this is the risky part, I don't think you want this but in that case we need to rename the class to make it clear it doesn't require the wheels but just needs a reference to it..

aschokking (changes requested):

Competition2017/src/competition/subsystems/shooter_belt/commands/RunBeltTracerSpeedModeCommand.java
imathno: We have both so we can use the one that seems to work more efficiently.

Competition2017/src/competition/subsystems/hybrid/commands/BaseShooterBeltAndWheelCommand.java
WasabiFan: Oh, interesting issue... I think we should just remove the new base class and put the `shooterWheelSubsystem` field directly into the command.

WasabiFan (changes requested):
I think we should remove the new base class and put the `shooterWheelSubsystem` field directly into the command. Then we can only require the one subsystem without semantics issues.

aschokking (approved):

WasabiFan (approved):

WasabiFan:
Can you make the agitator speed a separate property?

aschokking:
This looks reasonable, but we still don't have a good place to map it onto the OI for it to be used. Try and grab Andy to get his thoughts, I could imagine combining both commands into 1 button as well vs needing the left/right separation.

aschokking:
Need to figure out how to clear the common lib mix up too

PR #123, opened 2017-03-25
Moved trim direction to top of class file

Unagifan:
Moved TrimDirection enum to the top of class file. Changed default trim value to 100


Competition2017/src/competition/subsystems/shooter_wheel/commands/TrimShooterWheelCommand.java
```diff
@@ -10,19 +10,20 @@
 
     final DoubleProperty trimAmount;
     final ShooterWheelSubsystem shooterWheelSubsystem;
+    
     private TrimDirection direction;
     
+    public enum TrimDirection {
```
aschokking: Ideally I think this should be at the very top of the class, above the other instance variables

JohnGilb (approved):

PR #133, opened 2017-04-01
Invert shooter wheel to work with comp bot

imathno:


WasabiFan:
We need to make these inversions `BooleanProperty`s so that they're not dependent on hard-coded Booleans.

aschokking (approved):

aschokking:
The agreement with electrical was that the bots would be electrically identical so we wouldn't have to play that game (which has mostly panned out except for this right shooter). I'll put a task in the backlog to put properties in place next week for the practice bot.

PR #131, opened 2017-03-31
Fix autonomous to wait for shooter wheel to be at speed 

imathno:

PR #135, opened 2017-04-01
Replace hacky special-case check for nonzero shooter target speed

WasabiFan:
See https://github.com/Team488/TeamXbot2017/pull/131#discussion_r109279006

This is uglier than ideal because these commands take constructor parameters instead of being injected. The alternative isn't great either, so for now I'm leaving it as-is.


Competition2017/src/competition/subsystems/shooter_wheel/ShooterWheelSubsystem.java
```diff
@@ -102,12 +102,7 @@ public void setTargetSpeedForRange(double rangeInInches) {
     }
     
     public boolean isWheelAtSpeed() {
-        // Speed is in native units and can be negative
-        double currentSpeed = Math.abs(super.getSpeed());
-        if (currentSpeed <= 1) {
-            return false;
-        }
-        return currentSpeed >= super.getTargetSpeed() * wheelSpeedThresholdPercentage.get();
+        return Math.abs(getSpeed() - getTargetSpeed()) <= getTargetSpeed() * wheelSpeedThresholdPercentage.get();
```
imathno: Should we cast a absolute after the less then equal to?

aschokking: you mean if the targetspeed was negative for some reason?

aschokking: Looks reasonable, just to be safe let's make sure we test it out on the robot with a dry-run of autonomous in the pit / practice area before running this change in competition.

aschokking (approved):

Competition2017/src/competition/subsystems/shooter_wheel/ShooterWheelSubsystem.java
WasabiFan: Oh, good point!

WasabiFan:
I'd like to test this tonight and make sure that it works.

WasabiFan:
@aschokking Do you think it would be reasonable to merge this now and put an item on our to-do list to check the tuning? I want to get some mileage on this change.

aschokking:
Good question, given the current (albeit hacky) logic worked well through the last event I'm inclined not to add in a clean-up change that needs testing millage before all the other things that need it today (tracer, unjamming, auto, etc..). But if there ends up being extra time after everything else is vetted and settled to verify this I'm okay with that.

PR #137, opened 2017-04-04
Vision usability improvements

WasabiFan:


imathno (approved):

WasabiFan:
@aschokking Any objection to merging this? I don't think that there is anything here which could cause system-wide issues.

PR #139, opened 2017-04-05
Add more auto programs

tfrancisco:


Competition2017/src/competition/subsystems/autonomous/ShootAndDriveToHopperCommandGroup.java
```diff
@@ -0,0 +1,54 @@
...
+                180);
+        distanceToHopperFromTurningPoint = propManager.createPersistentProperty(
+                " distance from the turning point behind baseline to hopper in inches", 43.25);
+
+        this.shootAndBaseLine = shootAndBaseLine;
+        this.shootAndBaseLine.setAlliance(color);
+        this.addSequential(shootAndBaseLine);
+        
+        rotateToHopper = rotateToHopperProvider.get();
+        if(color == Alliance.Red) {
```
WasabiFan: Color isn't set by the time this runs; use the DS alliance

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/autonomous/ShootAndDriveToHopperCommandGroup.java
```diff
@@ -0,0 +1,54 @@
...
+    private final DoubleProperty blueAllianceHeadingToHopper;
+    private final DoubleProperty distanceToHopperFromTurningPoint;
+    
+    private Alliance color;
+
+    public ShootAndDriveToHopperCommandGroup(XPropertyManager propManager,
+            ShootAndDriveAcrossBaseLineCommandGroup shootAndBaseLine,
+            Provider<DriveForDistanceCommand> driveForDistanceProvider,
+            Provider<RotateToHeadingCommand> rotateToHopperProvider) {
+        redAllianceHeadingToHopper = propManager.createPersistentProperty(" red alliance heading to face hopper", 90);
```
WasabiFan: Remove preceding space and capitalize. Also, is there a reason we want this to be a property? I recall using a constant the last time we needed hopper headings.

Competition2017/src/competition/subsystems/autonomous/ShootAndDriveToHopperCommandGroup.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final DoubleProperty redAllianceHeadingToHopper;
+    private final DoubleProperty blueAllianceHeadingToHopper;
+    private final DoubleProperty distanceToHopperFromTurningPoint;
+    
+    private Alliance color;
+
+    public ShootAndDriveToHopperCommandGroup(XPropertyManager propManager,
+            ShootAndDriveAcrossBaseLineCommandGroup shootAndBaseLine,
+            Provider<DriveForDistanceCommand> driveForDistanceProvider,
```
WasabiFan: We can directly inject command instances if only one is needed

WasabiFan (dismissed):

WasabiFan:
I think I've finished this up. What I changed:
- The alliance should now be properly set
- I replaced the unused "drive to hopper then boiler" command button with this new one
- I made the new autonomous command sustain the shooter wheels after shooting finishes (so we can shoot immediately after auto). For this I created a subclass of the "stop all shooting" command which doesn't stop the shooters; the architecture feels backwards, but I couldn't find a better way to do it.

Also, I noticed that we have a bunch of `Setup<Name>Command`s which should really be replaced with a common setter class. I am not going to bother with those changes now, but in general we should aim for more concise structure.

imathno (approved):

Competition2017/src/competition/subsystems/autonomous/ShootAndActivateHopperCommandGroup.java
```diff
@@ -0,0 +1,65 @@
...
+
+        this.driveAcrossBaseline.setDeltaDistance(distanceFromShootingPositionToTurningPoint.get());
+        
+        rotateToHopper = rotateToHeadingProvider.get();
+        rotateToHopper.setTargetHeading(blueAllianceHeadingToHopper.get());
+        this.addSequential(rotateToHopper);
+        
+        driveToHopper = driveForDistanceProvider.get();
+        driveToHopper.setDeltaDistance(distanceToHopperFromTurningPoint);
+        this.addSequential(driveToHopper);
```
aschokking: we'll see when we test it but I was thinking we might want a little extra 'drive forward slowly for time' at the end to make sure the hopper is reached and pressed firmly.

Competition2017/src/competition/subsystems/shoot_fuel/StopFeedingAndCollectionCommandGroup.java
```diff
@@ -0,0 +1,21 @@
+package competition.subsystems.shoot_fuel;
+
+import com.google.inject.Inject;
+
+import competition.subsystems.agitator.AgitatorsManagerSubsystem;
+import competition.subsystems.collector.commands.StopCollectorCommand;
+import competition.subsystems.shooter_belt.ShooterBeltsManagerSubsystem;
+import competition.subsystems.shooter_wheel.ShooterWheelsManagerSubsystem;
+
+public class StopFeedingAndCollectionCommandGroup extends StopAllShootingCommandGroup {
```
aschokking, line 10: Consider inverting this relationship so the `StopAllShootingCommandGroup` extends `StopFeedingAndCollectionCommandGroup` by adding the shooterwheels into the mix?

aschokking: Ah I see now, you were using this to try and lie to the parent autonomous group that wanted a StopAllShootingCommandGroup, interesting..

Competition2017/src/competition/subsystems/autonomous/ShootAndActivateHopperCommandGroup.java
aschokking: it's probably prudent to have some kind of timeout on these drive commands in case the encoders go out so it doesn't go full power into the side

aschokking (approved):
Definitely need to test in practice matches, I expect we'll need to tune the distances quite a bit

Competition2017/src/competition/subsystems/shoot_fuel/StopFeedingAndCollectionCommandGroup.java
WasabiFan: Yeah, this felt particularly wrong while implementing it but I think it's the best structure that we can get right now. I needed to preserve the injectability of the "shoot and cross baseline" command, so this is what we end up with :grin:

WasabiFan:
This PR has been hijacked for broader autonomous changes.

Competition2017/src/competition/subsystems/autonomous/ShootFromHopperCommandGroup.java
```diff
@@ -0,0 +1,75 @@
...
+    private final DoubleProperty redHeadingToHopper;
+    private final DoubleProperty blueHeadingToHopper;
+
+    private final DoubleProperty distanceToTravelToHopper;
+    
+    protected final SetRobotHeadingCommand setInitialHeading;
+    protected final DriveForDistanceAtHeadingCommand activateHopper;
+    
+    @Inject
+    public ShootFromHopperCommandGroup(
```
JohnGilb: I would rename this to `ShootThenGoToHopperAndShoot` or something like that

WasabiFan: This command goes to the hopper, collects and shoots. I'm trying to differentiate this from "shoot and then align with hopper." What do you think?

Competition2017/src/competition/subsystems/autonomous/ShootAndDriveAcrossBaseLineCommandGroup.java
```diff
@@ -26,48 +25,51 @@
     
     private SetRobotHeadingCommand setInitialHeading;
     private ShootFuelForNSecondsCommandGroup shootFuel;
-    protected DriveForDistanceAtHeadingCommand driveAcrossBaseline;
```
WasabiFan: The heading-based command _is_ what we want here @imathno

Competition2017/src/competition/subsystems/autonomous/ShootFromHopperCommandGroup.java
```diff
@@ -0,0 +1,75 @@
...
+        this.setInitialHeading = setInitialHeading;
+        setInitialHeading.setHeadingToApply(blueAllianceStartingHeading.get()); 
+        addSequential(setInitialHeading);
+        
+        this.activateHopper = activateHopper;
+        activateHopper.setTargetHeadingProp(blueHeadingToHopper);
+        activateHopper.setTargetDistanceProp(distanceToTravelToHopper);
+        addSequential(activateHopper);
+        
+        // TODO: Rotate to boiler
```
JohnGilb: TODO: rotate and drive towards boiler

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+    public DriveForDistanceAtHeadingCommand(
+            DriveSubsystem driveSubsystem, 
+            XPropertyManager propMan,
+            PoseSubsystem pose,
+            RobotAssertionManager assertionManager,
+            PIDFactory pidFactory)
+    {
+        super(driveSubsystem);
+        this.poseSubsystem = pose;
+        speedHeadingThresh = propMan.createPersistentProperty("Drive at heading threshold for translation", 5);
```
JohnGilb: `speedHeadingThresh` was a little confusing to me - it's more verbose, but should this be something like `minHeadingErrorBeforeAllowingTranslation` ?

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+        
+        headingDrivePid = new PIDManager("Drive at heading rotation", propMan, assertionManager,  1 / 80d, 0, 0);
```
JohnGilb: Use the PIDFactory to do this. This also means you won't need the assertion manager.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+        
+        headingDrivePid.setErrorThreshold(3); // error in degrees, since degrees are used for the calculate call in execute()
```
JohnGilb: put these thresholds in the PIDFactory call I mentioned above, that way changes made via SmartDashboard will persist instead of being overridden by this.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
WasabiFan: I just inherited this. I would _rather_ not mess with it because the order of the parameters are different with the PID manager, and we don't have any way of proving that the new options work given that those properties already exist.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+            return;
+        }
+        
+        if(targetDistanceSupplier == null) {
+            log.error("DriveForDistanceAtHeading initialized without a distance setpoint!");
+            targetDistance = 0;
+            return;
+        }
+
+        targetHeading = new ContiguousHeading(targetHeadingSupplier.getAsDouble());
```
JohnGilb: I don't think ContiguousHeading can handle null.

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public void execute() {
+        if(targetHeading != null) {
+            double errorInDegrees = poseSubsystem.getCurrentHeading().difference(targetHeading);
+            double rotationalPower = headingDrivePid.calculate(0, errorInDegrees);
+            
+            
+            XYPair targetRelativePositionalError = poseSubsystem.getFieldOrientedTotalDistanceTraveled()
```
JohnGilb: I think you could replace this with `getDistanceToPoint` from XYPair class

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
```diff
@@ -0,0 +1,141 @@
...
+                    .add(initialPosition.scale(-1))
+                    .rotate(-targetHeading.getValue());
+            
+            double translationPowerFactor = Math.max(speedHeadingThresh.get() - Math.abs(errorInDegrees), 0) / speedHeadingThresh.get();
+            double translationPower = distancePid.calculate(targetDistance, targetRelativePositionalError.y) * translationPowerFactor;
+            
+            // Using the translation power as a multiplication factor ensures that we only align when we are moving toward the goal
+            double alignmentPower = alignmentPid.calculate(targetRelativePositionalError.x, 0) * Math.abs(translationPower);
```
JohnGilb: Since alignment is essentially another "rotation", it should share polarity with rotationalPower - right now it looks reversed.

Competition2017/src/competition/subsystems/shooter_wheel/ShooterWheelSubsystem.java
```diff
@@ -42,12 +46,18 @@ public ShooterWheelSubsystem(
...
         log.info("Creating");
         this.side = side;
+        
         flushToBoilerTargetSpeed = 
                 propManager.createPersistentProperty(side + " flush to boiler target speed", 9000);
         trimFlushToBoilerSpeed =
                 propManager.createEphemeralProperty(side + " trim speed", 0.0);
+        
+        offsetFromHopperTargetSpeed = 
+                propManager.createPersistentProperty(side + " offset from hopper target speed", 1100);
```
JohnGilb: should this default be something like 11,000 ?

JohnGilb (approved):

Competition2017/src/competition/subsystems/drive/commands/DriveForDistanceAtHeadingCommand.java
WasabiFan: Ugh, that's a good reason to change this legacy code... darn.

Competition2017/src/competition/subsystems/shooter_wheel/ShooterWheelSubsystem.java
WasabiFan: Good catch; this looks like a typo.

PR #143, opened 2017-04-12
Drive start/stop logging latch

aschokking:


Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -82,6 +88,20 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager, ShiftSub
         previousRightTicks = rightDrive.getPosition();
         leftInchesTraveled = 0;
         rightInchesTraveled = 0;
+        
+        drivingStateLatch.addObserver((Observable o, Object arg) -> {
```
WasabiFan: The types can be omitted

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -82,6 +88,20 @@ public DriveSubsystem(WPIFactory factory, XPropertyManager propManager, ShiftSub
...
+            if(arg instanceof EdgeType) {
+                EdgeType edge = (EdgeType)arg;
+                String type = "";
+                if(edge == EdgeType.RisingEdge) {
+                    type = "starting";
+                }
+                else if(edge == EdgeType.FallingEdge) {
+                    type = "stopping";
+                }
+                log.info(String.format("Drive %s left: %.2f, right: %.2f", type, this.leftDrive.get(), this.rightDrive.get()));
```
WasabiFan: What's the use of the raw values? This only runs when the state changes from below 0.01 to above that threshold, so these numbers will generally be one 0.01 and the other the arbitrary position that it was when the other started moving.

WasabiFan (changes requested):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: I agree that'll generally be the case for DriveWithJoysticks but for autonomous driving it seemed like it might be interesting to see what was being set in case things go whacky (even just the initial sign of things).

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
aschokking: Neat!

JohnGilb (approved):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
WasabiFan: I'm not sure what you mean. What is the value that printing the two powers in this location gives us? I could see a use case for periodic logging of power targets, but it seems to me that the numbers printed here will be of little use -- they'll distract more than they enlighten.

aschokking: I agree periodic logging would be nice too, I think I'd want this in addition to that but if you think it's net distracting to have the extra information there I'm happy to remove it since you'll be the one looking at it next week ;)

PR #146, opened 2017-04-15
set servo command and servo subsystem

tfrancisco:


Competition2017/src/competition/subsystems/servo/ServoSubsystem.java
```diff
@@ -0,0 +1,32 @@
...
+
+import xbot.common.command.BaseSubsystem;
+import xbot.common.controls.actuators.XServo;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class ServoSubsystem extends BaseSubsystem {
+    public XServo servo;
```
aschokking: final on these guys would help guard against booboos

Competition2017/src/competition/subsystems/servo/SetServoCommand.java
```diff
@@ -0,0 +1,30 @@
+package competition.subsystems.servo;
+
+import com.google.inject.Inject;
+
+import xbot.common.command.BaseCommand;
+import xbot.common.properties.XPropertyManager;
+
+public class SetServoCommand extends BaseCommand {
+    private ServoSubsystem servoSubsystem;
```
aschokking: final on these guys too

Competition2017/src/competition/subsystems/servo/SetServoCommand.java
```diff
@@ -0,0 +1,30 @@
...
+    private double targetPose;
+    
+    @Inject
+    public SetServoCommand(XPropertyManager propMan, ServoSubsystem servoSubsystem){
+        this.servoSubsystem = servoSubsystem;
+        this.requires(servoSubsystem);
+    }
+
+    @Override
+    public void initialize() {}
```
JohnGilb: `log.info("Initializing")`

Competition2017/src/competition/subsystems/servo/SetServoCommand.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    public void execute() {
+        servoSubsystem.setServo(targetPose);
+    }
+    
+    public void setTargetPose(double value) {
+        targetPose = value;
+    }
+
```
aschokking: I was thinking this command should end right away, or maybe after a short timer?

Competition2017/tests/competition/subsystems/servo/SetServoCommandTest.java
```diff
@@ -0,0 +1,22 @@
...
+import static org.junit.Assert.assertTrue;
+
+public class SetServoCommandTest extends BaseTest {
+    
+    @Test
+    public void test(){
+        SetServoCommand command = injector.getInstance(SetServoCommand.class);
+        ServoSubsystem subsystem = injector.getInstance(ServoSubsystem.class);
+        command.setTargetPose(0.4);
+        command.execute();
```
aschokking: generally a good idea to call initialize on your command here first in case it explodes/changes things

tfrancisco: would it be okay to make isFinished always return true since isFinished would be called after execute?

Competition2017/src/competition/subsystems/servo/SetServoCommand.java
tfrancisco: Why should it end right away?

Competition2017/src/competition/subsystems/servo/ServoSubsystem.java
```diff
@@ -0,0 +1,32 @@
...
+    private DoubleProperty minPose;
+    private DoubleProperty maxPose;
+    private double deltaPose;
+    
+    @Inject
+    public ServoSubsystem(WPIFactory factory, XPropertyManager propManager){
+        servo = factory.getServo(5);
+        
+        minPose = propManager.createPersistentProperty(" minimum pose for servo", 0.2);
```
WasabiFan: Remove preceding space, capitalize

Competition2017/src/competition/subsystems/servo/ServoSubsystem.java
```diff
@@ -0,0 +1,32 @@
...
+        maxPose = propManager.createPersistentProperty(" max pose for servo", 0.8);
+        deltaPose = maxPose.get() - minPose.get();
```
WasabiFan: I think we should calculate this in the `setServo` method so that it updates as the properties do.

Competition2017/src/competition/subsystems/servo/ServoSubsystem.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.inject.Singleton;
+
+import xbot.common.command.BaseSubsystem;
+import xbot.common.controls.actuators.XServo;
+import xbot.common.injection.wpi_factories.WPIFactory;
+import xbot.common.properties.DoubleProperty;
+import xbot.common.properties.XPropertyManager;
+
+@Singleton
+public class ServoSubsystem extends BaseSubsystem {
```
WasabiFan: I think we should specify what servos are being controlled. In this case, I'd recommend `ShooterAimAdjustmentSubsystem` and rename the packages and variables accordingly. You can choose though; the idea is that these servos aim the shooter pads a bit to extend the arc.

Competition2017/src/competition/subsystems/servo/ServoSubsystem.java
```diff
@@ -0,0 +1,32 @@
...
+    public XServo servo;
+    private DoubleProperty minPose;
+    private DoubleProperty maxPose;
+    private double deltaPose;
+    
+    @Inject
+    public ServoSubsystem(WPIFactory factory, XPropertyManager propManager){
+        servo = factory.getServo(5);
```
WasabiFan: There are actually two servos here: one on the left shooter and one on the right. Can you add a second servo which takes the same goal? It's worth noting that they are mirrored, so if one has a target of, say, `20%` the other should have a goal of `80%`. I am not sure what units the servo actually operates in -- presumably a percentage from 0 to 1.

WasabiFan (changes requested):
Looks pretty good! There are some small things that we need to update.

Competition2017/src/competition/subsystems/servo/SetServoCommand.java
WasabiFan: I think it would make sense to have it end immediately so it can be quickly chained into a `CommandGroup`. In my use case (which I believe is the only one), I want to tell the servo to move into the "engaged" position and continue doing other things in the autonomous program without delay.

WasabiFan: It would be reasonable to make it an option so that it can either run forever or end immediately.

WasabiFan:
Thanks for this @tfrancisco! There were some conflicts with other changes I made so I manually applied these changes (with some refactoring) to the autonomous branch that I'm working on.

PR #148, opened 2017-04-16
Add stop to drive subsystem

imathno:


Competition2017/src/competition/subsystems/drive/commands/StopDriveCommand.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        log.info("initializing");
+    }
+
+    @Override
+    public void execute() {
+        driveSubsystem.stop();
```
WasabiFan: Can we put this in `initialize` as well to ensure that it runs if `execute` isn't called before `isFinished`?

Competition2017/src/competition/subsystems/drive/commands/StopDriveCommand.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class StopDriveCommand extends BaseDriveCommand {
+
+    public StopDriveCommand(DriveSubsystem driveSubsystem) {
+        super(driveSubsystem);
+    }
+
+    @Override
+    public void initialize() {
+        log.info("initializing");
```
WasabiFan: Capitalize `I`

Competition2017/src/competition/subsystems/drive/commands/StopDriveCommand.java
```diff
@@ -0,0 +1,26 @@
+package competition.subsystems.drive.commands;
+
+import competition.subsystems.drive.DriveSubsystem;
+
+public class StopDriveCommand extends BaseDriveCommand {
+
+    public StopDriveCommand(DriveSubsystem driveSubsystem) {
```
WasabiFan: Ooh, final catch: this needs `@Inject`. Could you add a basic test which injects this command and confirms that it stops the motors? This should work, with some autocomplete mixed in where I can't remember method names:

```java
DriveSubsystem drive = // inject this
StopDriveCommand stopCommand = // inject this

drive.tankDrivePowerMode(0.5, 0.7);

// Assert that the powers were set -- there's probably a base test class for the drive which makes this easy

stopCommand.initialize();
assertTrue(stopCommand.isFinished());

// Assert that powers are zero
```

aschokking (approved):

Competition2017/src/competition/subsystems/drive/DriveSubsystem.java
```diff
@@ -206,6 +206,13 @@ public void tankDrivePowerMode(double leftPower, double rightPower) {
         updatePeriodicData();
     }
     
+    public void stop() {
+        leftDrive.set(0);
```
aschokking: should this make sure the talons are in vbus mode first?

aschokking: otherwise this could be setting a position goal of zero?

WasabiFan:
Thanks @imathno!

For future reference, the this targets an issue when a PID command exits but isn't exactly stopped or is terminated by a timeout; the 'bot keeps driving because there is no command to stop it. As a basic measure, I am going to add this command to the command group we are working on.

PR #149, opened 2017-04-21
Solenoid on the comp bot is inverted

imathno:


aschokking (approved):

WasabiFan:
We didn't really want to merge this; we manually made these changes in our working branch. It doesn't make much of a difference because we won't need to merge from master any more.

PR #150, opened 2017-07-29
Use new, simpler persistent property manager from CommonLib. 

JohnGilb:
Mostly this means removing Derby jars.


WasabiFan:
I think these changes need to be made in the template project as well.

